PR #4, opened 2015-04-07
Some work on KK module

frankhuster:


koshalt:
Bad merge. You're fired

PR #7, opened 2015-04-09
First KK-API checkin

frankhuster:
The API module now exposes a KK endpoint
New language module.
Works, but is not well documented.


language/src/main/resources/webapp/messages/messages.properties
```diff
@@ -0,0 +1,11 @@
+#Module name
+language=language
+
+#Buttons
+language.btn.sayHello=Say Hello
```
llavoie: There's still a little bit of Hello World junk here

language/src/main/resources/META-INF/motech/applicationContext.xml
```diff
@@ -0,0 +1,31 @@
...
+            <list>
+                <value>language</value>
+            </list>
+        </constructor-arg>
+        <constructor-arg name="i18n">
+            <map>
+                <entry key="messages" value="../language/resources/messages/" />
+            </map>
+        </constructor-arg>
+        <property name="defaultURL" value="/helloWorld"/>
```
llavoie: and here

Issue #9, opened 2015-04-09
api/kilkari - Get Inbox Details


llavoie:

larubbio:
I think we need a test for a valid subscriber with no inbox messages.  Or change our response if the subscriber is not found.  Currently we return 400.  I think we should instead return:

```javascript
{
  "inboxSubscriptionDetailList": []
}
```
According to the spec:

List of details of subscriptions having inbox.
If not present then IVR shall play promotional message.

Also the ivr checks like this:

```xml
   <if cond="lRespInboxDetailsObject.inboxSubscriptionDetailList.length &lt;= 0" >
      <log> No Messages Available </log>
      <goto next="#PromoMsg"/>
   </if>
```

frankhuster:
As of now, it looks like the code already returns an empty set:
```java
        Subscriber subscriber = kilkariService.getSubscriber(callingNumber);
        if (subscriber == null) {
            throw new NotFoundException(String.format(NOT_FOUND, "callingNumber"));
        }

        Set<Subscription> subscriptions = subscriber.getSubscriptions();
        Set<InboxSubscriptionDetailResponse> subscriptionDetails = new HashSet<>();
        for (Subscription subscription : subscriptions) {
            subscriptionDetails.add(new InboxSubscriptionDetailResponse(subscription.getSubscriptionId(),
                    subscription.getSubscriptionPack().getName(),
                    "10_1",
                    "xyz.wav"));
        }
```

Granted it's a bit hard-coded, but the spirit is right.
Should we close that until Sprint2?

larubbio:
It's the if check above.

If subscriber == null I think we should return. 200 with an empty
subscriptions list

frankhuster:
I'll look into it.
To tell @llavoie I'm taking over I'm assigning it to me, right?

frankhuster:
If ```kilkariService.getSubscriber(callingNumber)``` returns ```null``` it means there is no such subscriber in the database. It doesn't mean there is a valid subscriber with no subscription details. Am I missing something @larubbio ?

larubbio:
Correct, but if you look at the API spec this is the return values for
successful are:

```
{
    "circle": "AP",
    "languageLocationCode": “10”,
    "subscriptionPackList": [
        "48WeeksPack",
        "72WeeksPack"
    ]
}

Or

{
"circle": "AP",
    "defaultLanguageLocationCode": “10”
}

Or

{
"circle": "99",
    "defaultLanguageLocationCode": “10”
[...]

PR #30, opened 2015-04-14
Issue #9 - Get Inbox Details

llavoie:
Implemented the GetInboxDetails API in the API module, and some of the plumbing in Kilkari
(including adding a Subscription entity and refactoring existing Subscriber/SubscriptionPack
code a bit). The inboxWeekId and contentFileName are hard-coded to bogus values for now.

### DIFF
--- a/api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
+++ b/api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
@@ -1,5 +1,11 @@
-package org.motechproject.nms.kilkari.osgi;
+package org.motechproject.nms.api.osgi;
 
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.httpclient.HttpStatus;
+import org.apache.http.client.methods.HttpGet;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.motechproject.nms.kilkari.domain.Subscriber;
 import org.motechproject.nms.kilkari.domain.SubscriptionPack;
 import org.motechproject.nms.kilkari.domain.Subscription;
@@ -7,30 +13,32 @@ import org.motechproject.nms.kilkari.repository.SubscriberDataService;
 import org.motechproject.nms.kilkari.repository.SubscriptionPackDataService;
 import org.motechproject.nms.kilkari.repository.SubscriptionDataService;
 import org.motechproject.nms.kilkari.service.KilkariService;
-import org.junit.Test;
-import org.junit.runner.RunWith;
 import org.motechproject.testing.osgi.BasePaxIT;
 import org.motechproject.testing.osgi.container.MotechNativeTestContainerFactory;
+import org.motechproject.testing.osgi.http.SimpleHttpClient;
+import org.motechproject.testing.utils.TestContext;
 import org.ops4j.pax.exam.ExamFactory;
 import org.ops4j.pax.exam.junit.PaxExam;
 import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
 import org.ops4j.pax.exam.spi.reactors.PerSuite;
 
 import javax.inject.Inject;
-
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
 
 /**
- * Verify that KilkariService is present & functional.
+ * Verify that Kilkari API is functional.
  */
 @RunWith(PaxExam.class)
 @ExamReactorStrategy(PerSuite.class)
 @ExamFactory(MotechNativeTestContainerFactory.class)
-public class KilkariServiceBundleIT extends BasePaxIT {
+public class KilkariControllerBundleIT extends BasePaxIT {
+    private static final String ADMIN_USERNAME = "motech";
+    private static final String ADMIN_PASSWORD = "motech";
 
     @Inject
     private KilkariService kilkariService;
@@ -41,13 +49,7 @@ public class KilkariServiceBundleIT extends BasePaxIT {
     @Inject
     private SubscriptionDataService subscriptionDataService;
 
-    @Test
-    public void testServicePresent() throws Exception {
-        assertNotNull(kilkariService);
-    }
-
-    @Test
-    public void testServiceFunctional() throws Exception {
+    private void setupData() {
         subscriptionDataService.deleteAll();
         subscriptionPackDataService.deleteAll();
         subscriberDataService.deleteAll();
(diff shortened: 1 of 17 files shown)


api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -0,0 +1,85 @@
...
+
+        httpGet.addHeader("Authorization",
+                "Basic " + new String(Base64.encodeBase64((ADMIN_USERNAME + ":" + ADMIN_PASSWORD).getBytes())));
+
+        assertTrue(SimpleHttpClient.execHttpRequest(
+            httpGet,
+            "{\"inboxSubscriptionDetailList\":[{\"subscriptionId\":\"001\",\"subscriptionPack\":\"pack1\",\"inboxWeekId\":\"10_1\",\"contentFileName\":\"xyz.wav\"}]}"));
+    }
+
+}
```
larubbio: Should some negative test cases be included?  Like a callId without a subscriber?

api/src/test/java/org/motechproject/nms/api/osgi/UserControllerBundleIT.java
```diff
@@ -44,17 +46,25 @@
...
-        subscriberDataService.deleteAll();
-        Subscriber subscriber1 = subscriberDataService.create(new Subscriber("0000000000", onePack));
-        Subscriber subscriber2 = subscriberDataService.create(new Subscriber("0000000001", twoPacks));
+        Subscriber subscriber1 = subscriberDataService.create(new Subscriber("0000000000"));
+        Subscriber subscriber2 = subscriberDataService.create(new Subscriber("0000000001"));
+
+        Subscription subscription1 = subscriptionDataService.create(new Subscription("001", subscriber1, pack1));
+        Subscription subscription2 = subscriptionDataService.create(new Subscription("002", subscriber2, pack1));
+        Subscription subscription3 = subscriptionDataService.create(new Subscription("003", subscriber2, pack2));
     }
```
larubbio: Since these don't seem to be used in the tests are they needed as test data?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscriber.java
```diff
@@ -16,16 +18,17 @@
     @Unique
     private String callingNumber;
 
-    //todo: should this be a Set<> instead of a List<>?
+    //making this a bi-directional relationship until MOTECH-1638 is fixed
```
larubbio: Comment with TODO so the build and IDE flag it.  Also should the loading be set to lazy incase we forget to change it back to one direction?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -0,0 +1,73 @@
...
+
+import javax.jdo.annotations.Unique;
+
+@Entity
+public class Subscription {
+
+    @Field
+    @Unique
+    private String subscriptionId; //this is supposed to be a UUID, not enforcing that yet
+
```
larubbio: Mark with a todo so it is flagged in builds and IDE

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -0,0 +1,73 @@
...
+        return subscriptionId.equals(that.subscriptionId) && subscriber.equals(this.subscriber) && subscriptionPack.equals(that.subscriptionPack);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = subscriptionId != null ? subscriptionId.hashCode() : 0;
+        result = 31 * result + (subscriber != null ? subscriber.hashCode() : 0);
+        result = 31 * result + (subscriptionPack != null ? subscriptionPack.hashCode() : 0);
+        return result;
+    }
```
larubbio: Should this just return subscriptionId once it is a UUID?  If so should we mark this with a todo comment?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscriber.java
llavoie: Frank's checkin had the todo, mine removes it. I think he skirted the checkstyle warning by using lowercase todo rather than TODO :)

I'll change to lazy loading.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
llavoie: I created a separate bug for this already.

api/src/test/java/org/motechproject/nms/api/osgi/UserControllerBundleIT.java
llavoie: yeah, this was Frank's test data (I just rearranged it a bit). But I'll get rid of one of the subscribers since he's not using both.

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
llavoie: sigh. yes, they should.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
llavoie: the hash code, you mean?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/KilkariServiceImpl.java
```diff
@@ -20,12 +23,22 @@
...
+        }
+        return subscriptionPacksForSubscriber;
+    }
+
+    @Override
+    public Set<Subscription> getSubscriptions(String callingNumber) {
         Subscriber subscriber = subscriberDataService.findByCallingNumber(callingNumber);
         if (null == subscriber) {
             //todo: handle that better (ie: proper exception?) such that the proper bad response is returned
             throw new IllegalStateException("NULL subscriber");
```
larubbio: Instead of taking the callingNumber if this took the subcriber then the calling code can handle no subscriber for callingNumber.  Actually couldn't we just remove this function?  It doesn't seem to add that much value.

llavoie:
BTW, checkstyle complains when we put TODOs in the code -- so it might be better just to create bugs for things we want to fix later.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
larubbio: Yes.  I noticed you did something similar for subscriber (I think hashed
the number)

llavoie: ok -- adding a note about this to issue #32

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/KilkariServiceImpl.java
llavoie: yeah, I was thinking about just removing it and moving the retrieval of packs up to Frank's code...

larubbio:
I think it highlights them as a warning, but doesn't fail the build.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/KilkariServiceImpl.java
larubbio: I think that makes the most sense.  Doing it in here means we have to
handle the null case (and clearly we don't know how to).  Also all this
function really does is call subscriber.getSubscriptions() and the calling
code can do that themselves.

llavoie:
@larubbio This is ready for another look.

llavoie:
I'll make the change to check response bodies in next checkin -- created #44 so we don't forget. I'm going to merge this with Rob's permission.

PR #74, opened 2015-04-23
Fixes #67 Whitelist frontend functionality

### LINKED ISSUE #67: api/[ma,mk] If whitelisting is enabled compare user against it
The whitelist check occurs in getUser and setUserLanguage

larubbio:
This pull requests adds support for checking the whitelist at the api layer.  It also adds a minimal backend for the whitelist.


api/src/main/resources/META-INF/spring/blueprint.xml
```diff
@@ -17,6 +17,8 @@
```
larubbio: Are these osgi:reference tags needed?

llavoie: Does the code work without them? :)

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -38,10 +41,17 @@
 @Controller
 public class UserController extends BaseController {
 
+    public static final String CALLING_NUMBER = "callingNumber";
+    public static final String LANGUAGE_LOCATION_CODE = "languageLocationCode";
+    public static final String SERVICE_NAME = "serviceName";
```
koshalt: Are these the type of constants that we envision putting in the config/props package or is that just magic numbers and such?

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -80,17 +93,21 @@ public void setUserLanguage(@PathVariable String serviceName, @RequestBody Langu
 
         // TODO: If no FLW user is found we need to create one here #62
         FrontLineWorker flw = frontLineWorkerService.getByContactNumber(callingNumber);
-        if (null == flw) {
-            throw new NotFoundException(String.format(NOT_FOUND, "callingNumber"));
+        if (flw == null) {
+            throw new NotFoundException(String.format(NOT_FOUND, CALLING_NUMBER));
```
koshalt: Change this to NotImplementedException so that we don't miss it?

api/src/main/java/org/motechproject/nms/api/web/UserController.java
larubbio: I just did that to get checkstyle to shut up.  @fhuster gets testy when I
check in code that doesn't compile.

These are API level constants, so I think it might make sense to extract
all the json keys and put them in an api/json-keys.java file or something
like it.

api/src/main/java/org/motechproject/nms/api/web/UserController.java
larubbio: I have a test that depends on the values I set in there.  Ticket #62 was
put in to track this todo so no need for the NotImplementedException

When ticket #62 is done then those tests will fail (which is good, whoever
does #62 will then have to fix them for real)

koshalt: Would we still create the user if they aren't whitelisted? (like line 108)

flw/src/main/java/org/motechproject/nms/flw/domain/WhitelistEntry.java
```diff
@@ -0,0 +1,72 @@
...
+@Unique(name = "UNIQUE_STATE_CONTACT_NUMBER_COMPOSITE_IDX", members = { "state", "contactNumber" })
+public class WhitelistEntry {
+    public static final int FIELD_SIZE_10 = 10;
+
+    @Field
+    @NotNull
+    private State state;
+
+    @Field(required = true)
+    @Column(length = FIELD_SIZE_10)
```
koshalt: nitpick, but should we rename the field to something obvious like contactNumberLength or something? But I'm not should how the data migration would work if someone tried to modify the length of the column

flw/src/main/java/org/motechproject/nms/flw/domain/WhitelistEntry.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.validation.constraints.NotNull;
+
+@Entity(tableName = "whitelist_entries")
+@Unique(name = "UNIQUE_STATE_CONTACT_NUMBER_COMPOSITE_IDX", members = { "state", "contactNumber" })
+public class WhitelistEntry {
+    public static final int FIELD_SIZE_10 = 10;
+
+    @Field
+    @NotNull
+    private State state;
```
koshalt: Would the whitelist entry be a 'full' state or just a statecode and a contact number can then be looked up?

flw/src/main/java/org/motechproject/nms/flw/domain/WhitelistEntry.java
larubbio: I think that was also a checkstyle forced change that @fhuster made.  I
think it would make sense, but I think this is a case of checkstyle making
us do unnecessary work.  I think just putting 10 in the annotation is fine.

flw/src/main/java/org/motechproject/nms/flw/domain/WhitelistEntry.java
larubbio: I'm not sure I follow, but I think you have a good point.  We never really
load the whitelist entry, just check it's existence.  When users upload
data to populate it they will use the state code, not our state id.  (we
will need to validate on upload that we have a state for that state code
(I'll see if there is a jdo annotation that can set that foreign key))

location/src/main/java/org/motechproject/nms/location/service/impl/LocationServiceImpl.java
```diff
@@ -9,4 +11,20 @@
 @Service("locationService")
 public class LocationServiceImpl implements LocationService {
 
+    public static final int HACKY_LANGUAGE_CODE = 34;
+    public static final long HACKY_WHITELISTED_STATE_CODE = -1L;
+    public static final long HACKY_NON_WHITELISTED_STATE_CODE = -10L;
+
+    @Override
+    public State getStateForLanguage(Language language) {
+        // TODO: #71 For a given languageLocationCode find the districts it maps to and find the state for
```
koshalt: I thought that the whole point of language location code was that it would only map to one unique district? (vs. a language code that could map to districts across multiple states?)

larubbio: language location code maps to a set of districts, or to a state.

NMS.GEN.LANG.001

If there is only one “Language-location code” in the state, all districts
shall have same “Language-location code”; else, multiple groups shall be
formed, each group having a set of districts and having a unique
“Language-location code”.
Division or reorganization of geographical location for “Language-location
code” mapping is not supported and an item for future.

flw/src/main/java/org/motechproject/nms/flw/domain/WhitelistEntry.java
larubbio: It looks like I can't create a foreign key to a column other than id, so I can't use the stateCode here and still enforce referential integrity.  So instead this is going to remain as state_id, and when whitelist is uploaded we will load the state object and use it for the insert.

Issue #51, opened 2015-04-20
In API layer verify dates are in seconds since the epoch and then convert to DateTime


larubbio:

frankhuster:
Looks good.

PR #78, opened 2015-04-23
Fixes #72

### LINKED ISSUE #72: API: Ensure that our param data types match the spec
We'll need to take a pass through all of the endpoints in the API module and ensure that our param types match the spec. Right now we're accepting CallId and CallingNumber as Strings instead of Integers in most places, and this will break IMI. The validation code needs to be updated too.

frankhuster:


api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -24,11 +22,12 @@
     public static final String INVALID = "<%s: Invalid>";
     public static final String NOT_FOUND = "<%s: Not Found>";
 
-    public static final Pattern NUMERIC_PATTERN = Pattern.compile("[1-9][0-9]*|0");
```
larubbio: Do we need to enforce that it starts with a 1?  

Could you just use String regex = "\\d+";

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -24,11 +22,12 @@
...
-    public static final Pattern NUMERIC_PATTERN_10 = Pattern.compile("[1-9][0-9]{9}");
-    public static final Pattern NUMERIC_PATTERN_15 = Pattern.compile("[1-9][0-9]{14}");
+    public static final long SMALLEST_10_DIGIT_NUMBER = 1000000000L;
+    public static final long LARGEST_10_DIGIT_NUMBER  = 9999999999L;
+    public static final long SMALLEST_15_DIGIT_NUMBER = 100000000000000L;
+    public static final long LARGEST_15_DIGIT_NUMBER  = 999999999999999L;
 
-    protected static boolean validateFieldPresent(StringBuilder errors, String fieldName, String value) {
+    protected static boolean validateFieldPresent(StringBuilder errors, String fieldName, Object value) {
```
larubbio: Technically we are validating the field has a value.  I don't think we can actually determine the difference between these two json objects:

```javascript
{
   "bar": 1,
   "foo": null
}
```
and
```javascript
{
   "bar": 1
}
```

So I'm not sure we can every return a truly correct error.  I'm also not sure if they will get confused if we say it's not present when really it has no value.

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -36,62 +35,61 @@ protected static boolean validateFieldPresent(StringBuilder errors, String field
...
+        validateField15Digits(failureReasons, "callId", callId);
 
         return failureReasons;
     }
 
-    protected StringBuilder validate(String callingNumber, String operator, String circle, String callId) {
+    protected StringBuilder validate(Long callingNumber, Long callId, String operator, String circle) {
         StringBuilder failureReasons = validate(callingNumber, callId);
 
         validateFieldPresent(failureReasons, "operator", operator);
```
larubbio: According to the spec operator and circle are a max of 255 characters, so we should probably check that.

api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
```diff
@@ -75,7 +77,8 @@ public void saveCallDetails(@PathVariable String serviceName,
             throw new IllegalArgumentException(failureReasons.toString());
         }
 
-        FrontLineWorker flw = frontLineWorkerService.getByContactNumber(String.valueOf(callDetailRecordRequest.getCallingNumber()));
+        FrontLineWorker flw = frontLineWorkerService.getByContactNumber(
+                callDetailRecordRequest.getCallingNumber());
```
larubbio: Does this need to go on a new line once String.valueOf() was removed?

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
```diff
@@ -67,12 +67,12 @@ public InboxResponse getInboxDetails(@RequestParam String callingNumber, @Reques
...
     private StringBuilder validateSaveInboxCallDetails(InboxCallDetailsRequest request) {
-        StringBuilder failureReasons = validate(request.getCallingNumber(), request.getOperator(),
-                request.getCircle(), request.getCallId());
+        StringBuilder failureReasons = validate(request.getCallingNumber(), request.getCallId(),
+                request.getOperator(), request.getCircle());
 
-        validateFieldNumeric(failureReasons, "callStartTime", request.getCallStartTime());
-        validateFieldNumeric(failureReasons, "callEndTime", request.getCallEndTime());
-        validateFieldNumeric(failureReasons, "callDurationInPulses", request.getCallDurationInPulses());
+        validateFieldPresent(failureReasons, "callStartTime", request.getCallStartTime());
```
larubbio: We expect time related fields to be in seconds since the epoch, should we validate that and then convert to DateTime in the api layer?

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -70,16 +70,12 @@
     @ResponseStatus(HttpStatus.OK)
     public void setUserLanguageLocationCode(@PathVariable String serviceName,
                                             @RequestBody LanguageRequest languageRequest) {
-        String callingNumber = languageRequest.getCallingNumber();
-        String callId = languageRequest.getCallId();
-        String languageLocationCode = languageRequest.getLanguageLocationCode();
+        Long callingNumber = languageRequest.getCallingNumber();
+        Long callId = languageRequest.getCallId();
+        Integer languageLocationCode = languageRequest.getLanguageLocationCode();
```
larubbio: Ticket #79 is about converting languageLocationCode and defaultLanguageLocation code to Strings.  Should we do that here or in a different commit?

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
frankhuster: Sure, so our service layer uses DateTime instead of seconds since epoch.

frankhuster:
@larubbio want me to take over & merge?

larubbio:
@fhuster I've got it.  I completed the merge and now I'm finishing the
review.

api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
larubbio: It's 3 characters over, but I'm putting it back on one line anyway.

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
larubbio: Ticket #51 is tracking this.

PR #95, opened 2015-04-25
More/better comments - fixes #92

### LINKED ISSUE #92: Harmonize class names for request & response DTOs
If there was a 'low priority' label I would have tagged this issue with it.

frankhuster:
Busy work...
Changing the title didn't seem to do it, so I'm adding the magic words below.
Fixes #92


api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
```diff
@@ -78,7 +76,7 @@ public InboxResponse getInboxDetails(@RequestParam(required = false) Long callin
     }
 
     private void validateSaveInboxCallDetailsContent(StringBuilder failureReasons,
-                                                     Set<InboxCallDetailsRequestCallData> content) {
+                                                     Set<CallDataRequest> content) {
```
koshalt: There isn't a CallData request per se, right? CallData object is just part of a list inside of CallDetailsRequest?

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -143,7 +142,7 @@ private ResponseUser getFrontLineWorkerResponseUser(String serviceName, Long cal
             service = Service.MOBILE_KUNJI;
         }
 
-        user = new FrontLineWorkerUser();
+        user = new FlwUserResponse();
```
koshalt: Just curious, why is this being split from the declaration in line 133?

koshalt:
Builds and tests fine. The comments shouldn't be blocking and feel free to merge once you get back

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
frankhuster: Yep

api/src/main/java/org/motechproject/nms/api/web/UserController.java
frankhuster: Don't know. But good point, I fixed it.

frankhuster:
@koshalt blessed this change, merging it.

Issue #27, opened 2015-04-09
The return format for new and existing users will be combined


larubbio:
The VXML determines a new user based on languageLocationCode being set to a value.  They ignore defaultLanguageLocationCode.  To simplify the system the returned format for new and existing users are being combined.

Issue #26, opened 2015-04-09
We will no longer return circle in the user apis


larubbio:
In these APIs circle is passed to us so there is no need to return it.

PR #109, opened 2015-04-25
Fixes #16

### LINKED ISSUE #16: api/kilkari - Get Subscriber Details

frankhuster:
To the best of my latest understanding of the spec / spec changes
Also, noticed that languageLocationCode was missing from the request
parameters of /kilkari/inbox (see 4.2.2.1.3) so I added it.

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/UserController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/UserController.java
@@ -116,25 +116,21 @@ public class UserController extends BaseController {
     private UserResponse getKilkariResponseUser(Long callingNumber, String circle) {
         KilkariUserResponse user = new KilkariUserResponse();
         Set<String> packs = new HashSet<>();
+
         Subscriber subscriber = kilkariService.getSubscriber(callingNumber);
         if (subscriber != null) {
             Set<Subscription> subscriptions = subscriber.getSubscriptions();
             for (Subscription subscription : subscriptions) {
                 packs.add(subscription.getSubscriptionPack().getName());
             }
-        } else {
-            /*
-            No subscriber found in the database, according to
-            https://github.com/koshalt/mim/issues/9#event-288912397
-            we need to respond with an empty subscription list, the circle provider in the request and our
-            guess as to what the defaultLanguageLocationCode is, which is done after this method returns
-            */
-            Language language = languageService.getDefaultCircleLanguage(circle);
-            if (language != null) {
-                user.setDefaultLanguageLocationCode(language.getCode());
-            }
         }
         user.setSubscriptionPackList(packs);
+
+        Language language = languageService.getDefaultCircleLanguage(circle);
+        if (language != null) {
+            user.setDefaultLanguageLocationCode(language.getCode());
+        }
+
         return user;
     }
 
--- a/api/src/main/java/org/motechproject/nms/api/web/UserController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/UserController.java
@@ -98,7 +98,7 @@ public class UserController extends BaseController {
         Kilkari in the house!
          */
         if (KILKARI.equals(serviceName)) {
-            user = getKilkariResponseUser(callingNumber, circle);
+            user = getKilkariResponseUser(callingNumber);
         }
 
         if (failureReasons.length() > 0) {
@@ -113,7 +113,7 @@ public class UserController extends BaseController {
         return user;
     }
 
-    private UserResponse getKilkariResponseUser(Long callingNumber, String circle) {
+    private UserResponse getKilkariResponseUser(Long callingNumber) {
         KilkariUserResponse user = new KilkariUserResponse();
         Set<String> packs = new HashSet<>();
         Subscriber subscriber = kilkariService.getSubscriber(callingNumber);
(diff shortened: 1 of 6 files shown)


larubbio:
I think including languageLocationCode is another case were the spec is wrong.  The vxml isn't actually sending that parameter over, so we can drop it and put a ticket in to update the spec:

```xml
   <data name="lRespInboxDetails" src="http://172.16.96.211:8080/motech-platform-server/module/kilkari/inbox?" namelist="callingNumber callId" enctype="application/json" />
```

frankhuster:
But, it was just **added** to the spec!
:confused:

larubbio:
When was it added to the spec?  I think it has always been there.

frankhuster:
It's in yellow in the PDF:

yellow-languagelocationcode

larubbio:
Interesting.  It isn't in v1.1 but it is in v1.2.  However the summary of
changes they asked me to review makes no mention of it.  I still think it
is an error since MOTECH tells the ivr the language code (unless the IVR is
collecting it from the user)  I will email them for more information.

frankhuster:
@larubbio want me to back out the languageLocationCode changes and keep the nonexistent subscriber logic?

larubbio:
I think so.  Of course that means it is almost certain that the response
will be that it is supposed to be there. :)

frankhuster:
OK, languageLocationCode removed.

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -114,16 +113,24 @@ public UserResponse getUserDetails(@PathVariable String serviceName,
...
+            }
+        } else {
+            /*
+            No subscriber found in the database, according to https://github.com/koshalt/mim/issues/9#event-288912397
+            we need to respond with an empty subscription list, the circle provider in the request and our guess as
+            to what the defaultLanguageLocationCode is, which is done after this method returns
+            */
+            user.setCircle(circle);
+            //todo: figure out the defaultLanguageLocationCode, should that be coming from the language module?
+            user.setDefaultLanguageLocationCode("??");
```
larubbio: Is this if block needed?  According to issue #26 We aren't returning circle anymore.  Also according to issue #27 We are combining the return for new and existing.  So just send back both languageLocationCode and defaultLanguageLocationCode.  Also you can find the default language the same we we find it in MA/MK (https://github.com/motech-implementations/mim/blob/master/api/src/main/java/org/motechproject/nms/api/web/UserController.java#L109)

api/src/main/java/org/motechproject/nms/api/web/contract/kilkari/KilkariUserResponse.java
```diff
@@ -14,6 +14,7 @@
  */
 public class KilkariUserResponse extends UserResponse {
     private Set<String> subscriptionPackList;
+    private String circle;
```
larubbio: We don't need to return this, so it can go away.

api/src/test/java/org/motechproject/nms/api/osgi/UserControllerBundleIT.java
```diff
@@ -332,7 +332,7 @@ private HttpPost createHttpPost(String service, UserLanguageRequest request) thr
         return httpPost;
     }
 
-    private String createKilkariUserResponseJson(String defaultLanguageLocationCode, String locationCode,
+    private String createKilkariUserResponseJson(String defaultLanguageLocationCode, String locationCode, String circle,
```
larubbio: circle can be removed from here as well.

frankhuster:
Err, so should I not bother reading the spec?

:confused: 

circle

larubbio:
But you were the one who said we shouldn't return parameters that are sent
to us.  That's why we added the ticket to amend the spec.

frankhuster:
Stripped #16 now down to the barebones. Should be an easy code review, anybody.

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -116,14 +115,21 @@ public UserResponse getUserDetails(@PathVariable String serviceName,
...
         Set<String> packs = new HashSet<>();
-        for (Subscription subscription : subscriptions) {
-            packs.add(subscription.getSubscriptionPack().getName());
+        Subscriber subscriber = kilkariService.getSubscriber(callingNumber);
+        if (subscriber != null) {
+            Set<Subscription> subscriptions = subscriber.getSubscriptions();
+            for (Subscription subscription : subscriptions) {
+                packs.add(subscription.getSubscriptionPack().getName());
+            }
+        } else {
```
larubbio: You can delete this else statement and always set and return defaultLanguage

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -116,14 +115,21 @@ public UserResponse getUserDetails(@PathVariable String serviceName,
...
+            /*
+            No subscriber found in the database, according to https://github.com/koshalt/mim/issues/9#event-288912397
+            we need to respond with an empty subscription list, the circle provider in the request and our guess as
+            to what the defaultLanguageLocationCode is, which is done after this method returns
+            */
+            //todo: figure out the defaultLanguageLocationCode, should that be coming from the language module?
+            user.setDefaultLanguageLocationCode("??");
         }
         user.setSubscriptionPackList(packs);
```
larubbio: This is how we get defaultLanguage in UserController.  I would do the same here.

```java
        Language language = languageService.getDefaultCircleLanguage(circle);
        if (language != null && user != null) {
            user.setDefaultLanguageLocationCode(language.getCode());
        }
```

frankhuster: Darn, forgot to include that.
...coming right up!

larubbio:
@frankhuster Do you want to review?

PR #191, opened 2015-04-27
Partial fix for #13

llavoie:
Adds new OBD module and API for CDR Notification. Needs more negative test cases for the parameter validation. Also adds stub for File Processed Status API (#14) which will be filled in on a subsequent checkin.


outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/web/OutboundDialerController.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * OutboundDialerController
+ */
+@Controller
+public class OutboundDialerController {
+
+    public static final String NOT_PRESENT = "<%s: Not Present>";
+    public static final String INVALID = "<%s: Invalid>";
+    public static final Pattern TARGET_FILENAME_PATTERN = Pattern.compile("OBD_NMS[1-9]_20[0-9]{12}\\.csv");
```
frankhuster: We (MOTECH) generate target files and 'send' them to the IVR system.
We receive CDR files (summary & detailed) from the IVR syatsm. So this should be either CDR_SUMMARY_FILENAME_PATTERN or CDR_DETAILED_FILENAME_PATTERN and check conformity with point 14 of section 4.1.5 in the spec. (ie: CDR_Summary_xxx or CDR_Detail_xxx)

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/web/OutboundDialerController.java
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * IVR shall invoke this API to notify when a CDR file is ready.
+     */
+    @RequestMapping(value = "/cdrFileNotification",
+            method = RequestMethod.POST,
+            headers = { "Content-type=application/json" })
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public void notifyNewCdrFile(@RequestBody CdrFileNotificationRequest cdrFileNotificationRequest) {
+        StringBuilder failureReasons = new StringBuilder();
+        validateTargetFileName(failureReasons, cdrFileNotificationRequest.getFileName());
```
frankhuster: 4.2.6.1.4 mentions the following sample body
```json
{
   "fileName": "OBD_NMS1_20150127090000.csv",
   "cdrSummary":
   {
      "cdrFile": "cdrSummary_OBD_NMS1_20150201090000.csv",
      "checksum": "xxxxxx",
      "recordsCount": 5000
   },
   "cdrDetail":
   {
      "cdrFile": "cdrDetail_OBD_NMS1_20150201090000.csv",
      "checksum" : "xxxxxx",
      "recordsCount": 9900
   }
}
```
`fileName` is what we passed to the IVR system originally, shouldn't we be validating `cdrSummary` & `cdrDetail` instead? (or in addition?)

frankhuster:
Couple of things I think should be done in your next checkin.

frankhuster:
Hey @llavoie I'm not deleting this branch, since I assume you'll still be working on it.
But you're welcome to delete it yourself and start a new one with the closest issue # you think.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/web/OutboundDialerController.java
llavoie: We do validate all 3 names. I use a regex for the target file pattern, and then check that the other two files are equivalent to the target file with 'cdrSummary_' or 'cdrDetail_ appended at the beginning.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/web/OutboundDialerController.java
llavoie: We do validate all three. See other comment.

frankhuster: Doh!
Must have vision checked...

frankhuster:
@llavoie I've been wondering.
Why shouldn't we put the outbound-dialer http endpoints with in the api module?

larubbio:
We discussed this in the office.  One of the spec req. is that we can
change the ivr provider without changing the services code.  If we put the
obd code in the services code then we break that.

frankhuster:
I'm just talking about the http endpoints. 
The OBD service code can still be in the OBD module.
Having the api part in the api module actually decouples the api from the service.
I must be misunderstanding something...

larubbio:
But these apis are specific to this outbound dialer.  Wouldn't we want
those together?

frankhuster:
Not sure. I see some code duplication (in the request/param verification & error handling) which I remember @llavoie told me about, which is not a big deal. 

More to the point, it feels like the functionality of checking requests, calling service apis & returning responses fits really well in the api module. This is where all nms modules (including obd) interface with the world.

I feel more strongly about this than I do the obd and its http endpoints being isolated. 

It seems the (non)dependency on the IVR system is similar in both cases anyway. Right?

But it's not a super strong feeling.

larubbio:
When we discussed this we all agreed that the functionality of checking
requests does naturally fit together, however we also decided that there
was a difference between the business services of KK, MA and MK and the IVR
interaction specific to IMI.  That was why we pulled it out.

The api package contains the code that integrates with the VXML.  The obd
platform contains the code that integrates with IMI, more specifically with
the DVP system.  If we move to a different IVR vendor the api module will
not need to change.  The obd module will.  That is the separation we are
going for.

frankhuster:
Got it!

PR #192, opened 2015-04-27
Fixes #98 - Add domain classes and simple tests for the location hierarchy

### LINKED ISSUE #98: NMS.GEN.LOC.002 - Location Database
The NMS system shall manage a location database having the following parameters to manage the location of an FLW for MA/MK service and a Beneficiary for Kilkari:
 - State
 - District
 - Taluka
 - Health block
 - Health Facility (PHC/CHC)
 - Sub center
 - Village

There shall be mapping between descriptive name and codes.
The exact definitions of these fields along with presence (mandatory/optional) are provided in Appendix A (refer section 6.2).

The location data is used to link a given FLW record with a particular geographical location. There is a one to one link from a FLW record to a location record.

MCTS follows proposed census codes for all the location fields except Health block, Facility and Sub center. For Health block, facility and sub center MCTS has its own codes.

Villages are of two types – Census villages and Non census villages. For Non census villages, data entry person has option to enter the village name. The NMS system will only have one field for Village – regardless of whether it’s a census village, Panchayat village, revenue village or otherwise.

larubbio:


location/src/main/java/org/motechproject/nms/location/domain/HealthFacility.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "HealthFacility{" +
+                "name='" + name + '\'' +
+                ", code=" + code +
+                ", healthFacilityType=" + healthFacilityType +
+                ", healthBlock=" + healthBlock +
+                '}';
```
frankhuster: No healthSubFacilities?

location/src/main/java/org/motechproject/nms/location/domain/State.java
```diff
@@ -1,41 +1,47 @@
...
     @UIDisplayable(position = 1)
-    private Long stateCode;
+    @NotNull
+    private Long code;
 
     @Field
+    @Cascade(delete = true)
     @Persistent(mappedBy = "state", defaultFetchGroup = "true")
-    private Set<District> districts;
+    private List<District> districts;
```
frankhuster: Why the change from Set to List?

location/src/main/java/org/motechproject/nms/location/domain/Taluka.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Field
+    @UIDisplayable(position = 2)
+    @NotNull
+    private District district;
+
+    @Field
+    @Cascade(delete = true)
+    @Persistent(mappedBy = "taluka", defaultFetchGroup = "true")
+    private List<HealthBlock> healthBlocks;
```
frankhuster: Collections of children entities are not Set anymore?

location/src/main/java/org/motechproject/nms/location/domain/HealthFacility.java
larubbio: I was using the rule that the equals only looked at primitive fields in the
object and the toString looked at the object and it's parent.

location/src/main/java/org/motechproject/nms/location/domain/Taluka.java
larubbio: Kosh and I decided that Set is hard to work with.  You can't just .get()
from it.

frankhuster: Agreed, so you're cool with the fact that equality won't not be guaranteed when comparing similar Lists when it would be guaranteed for Sets?

location/src/main/java/org/motechproject/nms/location/domain/HealthFacility.java
frankhuster: Why?

frankhuster:
It's weird, I checked out this branch and nearly all the `UserControllerBundleIT` tests fail on my machine with something like:
```
testGetUserDetailsServiceCapped(org.motechproject.nms.api.osgi.UserControllerBundleIT)  Time elapsed: 0.021 sec  <<< ERROR!
com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails (`motech_data_services`.`FLW_MODULE_FRONTLINEWORKER`, CONSTRAINT `FLW_MODULE_FRONTLINEWORKER_FK2` FOREIGN KEY (`district_id_OID`) REFERENCES `nms_districts` (`id`))
```
It looks like some database state stuff. Confused why the CI didn't also fail...

larubbio:
You have existing data but the ci does not.  It means the delete needs to
delete more than it is.

location/src/main/java/org/motechproject/nms/location/domain/HealthFacility.java
larubbio: Why what?

frankhuster: Why should we be excluding collection data members from equals?

frankhuster:
But doesn't the CI do a `drop database`?
See [this section of the parent pom.xml](https://github.com/motech-implementations/mim/blob/master/pom.xml#L1084).

location/src/main/java/org/motechproject/nms/location/domain/HealthFacility.java
larubbio: It was causing trouble.  I was getting stack overflow and detached member
exceptions.  Also the code is unique in the database so I was thinking that
is god for equality testing.

larubbio:
That is probably why it had less data than you do locally

frankhuster:
I ran the IT on my machine: eeek 
See  http://pastebin.com/4AwnpNt0
:confused:

larubbio:
Odd, this change shouldn't affect that.

That error is from trying to delete an FLW but there is still a district in
the database.  If that district was connected to a state (i.e. one that was
created in the location test) then the stateDataService.deleteAll() would
clean it up.  I'll still look into it.

location/src/main/java/org/motechproject/nms/location/domain/Taluka.java
larubbio: We thought the simpler way of working with lists (i.e. .get() as opposed to converting to an array and then grabbing elements) was worth the tradeoff.  If there was something were we need to compare equality across the elements then set probably makes sense.  In this case we aren't comparing them in that way.

larubbio:
@frankhuster @koshalt @llavoie Anyone want to wrap up this review?

location/src/main/java/org/motechproject/nms/location/domain/Taluka.java
frankhuster: Converting to an array isn't necessary, but still it's a bit of a pain to get a random element:
```java
for (Iterator<Foo> it = set.iterator(); it.hasNext(); ) {
   return it.next();
}
```

Issue #167, opened 2015-04-26
NMS.KK.ACCESS.015 - Rules for managing subscription state


larubbio:
Kilkari shall manage the subscription state as below:
- **Active**: Kilkari shall set the status of the subscription to “Active” on the start date of the subscribed pack.
- **Completed**: Kilkari shall set the status of the subscription to “Completed” after delivery (including retries if any) of the last message of the subscribed pack.
- **Deactivated**: Kilkari shall set the status of the subscription to “Deactivated” after a beneficiary has chosen the option to deactivate or due to other reasons ( #162  NMS.KK.ACCESS.010, #169 NMS.KK.OBD.008 & #170  NMS.KK.DND.002)
- **Pending Activation**: Kilkari shall set the status of the subscription to “Pending Activation” if the start date of the pack is in future.

larubbio:
![screen shot 2015-04-27 at 10 27 30 am](https://cloud.githubusercontent.com/assets/51527/7353446/792d6d96-ecc8-11e4-9fcf-46332587bed5.png)

frankhuster:
#243 takes care of #166 and the **Completed** state portion of this issue.

llavoie:
I think the only part of this that isn't addressed yet is the transition from PENDING_ACTIVATION to ACTIVE.

llavoie:
Should we have a SubscriptionActivator (similar to SubscriptionDeactivator) that is responsible for transitioning subscriptions from PENDING to ACTIVE once the subscriber reaches LMP + 90 days?

llavoie:
We could add this to the OBD logic -- we'd need to change the target file generation to get all active and pending subscriptions due for a day (for now it only gets active), and update the pending ones to be active.

Issue #157, opened 2015-04-26
NMS.KK.ACCESS.004 - Subscription start date following MCTS upload


larubbio:
If the subscriber does NOT call the IVR long code/s, and is subscribed automatically from the back end via an upload of data from MCTS, then the subscription pack for a beneficiary and the week that the message will start in will be based on the LMP or DOB provided in MCTS Data upon data upload (refer #110 NMS.GEN.MCTS.001) as per following rules:
- If DOB is present (with or without LMP) in the data, then Child pack (12-Months-Pack) shall be subscribed. Service shall be “activated” from the date of data upload (subscription state = Active).
- If LMP is present without DOB then pregnancy pack shall be subscribed. Service shall be “activated” from the date of data upload (subscription state = Active) except for case of “early subscription” (for which refer #160 NMS.KK.ACCESS.007).

If a woman is subscribed to Kilkari via MCTS data, the service could start at any point during the 72 week or 48 week schedule. Since we cannot assume that her ASHA or ANM will have informed her about Kilkari, we will need to play her an initial welcome message before any other content is played. This welcome message will introduce her to Kilkari. Once this welcome message has been played in the first week of the service starting, then the next week’s content, as per the LMP or DOB, will be played.

larubbio:
![screen shot 2015-04-27 at 10 27 30 am](https://cloud.githubusercontent.com/assets/51527/7354144/e21bab2e-eccd-11e4-934f-9206d1dde1df.png)

larubbio:
From an email thread:

    Handling for following requirement is missing in API specification. 
    
    "If a woman is subscribed to Kilkari via MCTS data, the service could start at any point during the 72
    week or 48 week schedule. Since we cannot assume that her ASHA or ANM will have informed her
    about Kilkari, we will need to play her an initial welcome message before any other content is played."

    To handle this requirement a new field is needed in OBD record (in OBD file) to indicate the first 
    OBD message of subscribed pack.

llavoie:
@larubbio Could you ask for additional clarification on the email thread you mention above?

There are a few more things I think we need to know:

- What do we add to the OBD target file to indicate that the welcome message needs to be played? Is this a new column, or a weekId/filename that indicates the welcome message?
- Do we need to indicate the same in response to an inbox request?
- How do we set the start date for a pregnancy pack when we have an LMP? As I understand it, the pregnancy pack starts at the second trimester. So is the start date something like LMP + 90ish days?

[...]

Issue #160, opened 2015-04-26
NMS.KK.ACCESS.007 - "Early Subscription" definition and rules


larubbio:
Kilkari shall support “Early subscription” to Pregnancy pack only via MCTS bulk upload (subscription state = Pending Activation).

“Early subscription” refers to the case where Kilkari OBD message delivery for pregnancy pack shall be started from 2nd trimester (computed from LMP date), even though the MCTS record is uploaded before the actual date of 2nd trimester’s start.

larubbio:
![screen shot 2015-04-27 at 10 27 30 am](https://cloud.githubusercontent.com/assets/51527/7355003/e3beaa98-ecd3-11e4-8413-7d6607209b6c.png)

PR #198, opened 2015-04-29
Fixes #156, fixes #158, fixes #159

### LINKED ISSUE #156: NMS.KK.ACCESS.002 - Kilkari activation via IVR
Kilkari Services for a beneficiary can be subscribed via IVR Long code. There shall be separate long codes for Each pack.

Service shall be “activated” from the date of IVR call (subscription state = Active).

Exact long code shall be finalized during deployment. 

Subscription via IVR would imply that the location and other vital information of the beneficiary shall not be available in the NMS (only MSISDN and LMP/DOB shall be determined based on date of call) such that subscription for various packs start from next day.

IVR system shall identify the circle that the user is calling from (from their mobile number) in order to identify their language preference. If there are multiple languages available in the circle (North East for example) the user will be prompted to enter a language code, obtained from their ASHA, to set the language. The default prompt will be in the default language for their circle, or Hindi if circle is not known. Refer section 3.1.4 for details on language selection.

### LINKED ISSUE #158: NMS.KK.ACCESS.005 - Rules for multiple subscriptions for a single MSISDN
Kilkari shall allow subscription of service for different packs on the same MSISDN (via bulk upload of MCTS data or via IVR). 

For same pack, if an old subscription has completed/deactivated, then an additional subscription can be made.

MCTS data records with same MSISDN with one having DOB and other having only LMP shall be allowed to subscribe to different packs.

### LINKED ISSUE #159: NMS.KK.ACCESS.006 - Reason to reject a subscription (more multiple MSISDN subscription rules)
Kilkari Service shall not allow multiple “Active” or “Pending Activation” subscriptions for the same pack on the same MSISDN via bulk upload of MCTS Data or via IVR. In particular, subscription request with a particular MSISDN, shall be rejected IF: 
- DOB (with or without LMP) is present in the MCTS record and there is an already “Active” subscription to Child Pack on this MSISDN
- LMP (without DOB) is present in the MCTS record and there is an already “Active” or “Pending Activation” subscription to Pregnancy Pack on this MSISDN.
- IVR Subscription is requested and there is an already “Active” or “Pending Activation” subscription to that Pack on this MSISDN.
If Subscription for a pack is completed or gets deactivated, then only new subscription for the 

In case of MCTS bulk upload, rejected record should be logged.

In case of IVR, the subscription failure message should be played to the user.

llavoie:
Kilkari subscription state machine first phase (there's still a bit more work to do here, e.g. #167, #157, etc.)


api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -212,22 +220,145 @@ public void testCreateSubscriptionRequest() throws IOException, InterruptedExcep
...
+        mctsSubscriber.setDateOfBirth(LocalDate.now().minusDays(14));
+        subscriberDataService.create(mctsSubscriber);
+
+        subscriptionService.createSubscription(9999911122L, "10", "pack1", SubscriptionMode.MCTS_IMPORT);
+
+        // attempt to create subscription to the same pack -- should fail
+        subscriptionService.createSubscription(9999911122L, "10", "pack1", SubscriptionMode.MCTS_IMPORT);
+
+        mctsSubscriber = subscriberDataService.findByCallingNumber(9999911122L);
+        assertEquals(1, mctsSubscriber.getSubscriptions().size());
```
larubbio: Should this assert also be earlier, just to make sure it is the second call that fails and not the first.

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -212,22 +220,145 @@ public void testCreateSubscriptionRequest() throws IOException, InterruptedExcep
...
+        mctsSubscriber.setLastMenstrualPeriod(LocalDate.now().minusDays(28));
+        subscriberDataService.create(mctsSubscriber);
+
+        subscriptionService.createSubscription(9999911122L, "10", "pack2", SubscriptionMode.MCTS_IMPORT);
+
+        // attempt to create subscription to the same pack -- should fail
+        subscriptionService.createSubscription(9999911122L, "10", "pack2", SubscriptionMode.MCTS_IMPORT);
+
+        mctsSubscriber = subscriberDataService.findByCallingNumber(9999911122L);
+        assertEquals(1, mctsSubscriber.getSubscriptions().size());
```
larubbio: Same as above comment

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -212,22 +220,145 @@ public void testCreateSubscriptionRequest() throws IOException, InterruptedExcep
...
+        mctsSubscriber = subscriberDataService.findByCallingNumber(9999911122L);
+        Subscription pregnancySubscription = mctsSubscriber.getSubscriptions().iterator().next();
+        pregnancySubscription.setStatus(SubscriptionStatus.DEACTIVATED);
+        subscriptionDataService.update(pregnancySubscription);
+
+        // attempt to create subscription to the same pack -- should succeed
+        subscriptionService.createSubscription(9999911122L, "10", "pack2", SubscriptionMode.MCTS_IMPORT);
+
+        mctsSubscriber = subscriberDataService.findByCallingNumber(9999911122L);
+        assertEquals(2, mctsSubscriber.getSubscriptions().size());
```
larubbio: I'm a little concerned that we might return deactivated subscriptions back to the IVR.  Should we split getSubscriptions into getActiveSubscriptions and getAllSubscriptions?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/SubscriptionPack.java
```diff
@@ -14,13 +13,19 @@
...
-    public SubscriptionPack(String name) {
+    @Field
+    private SubscriptionPackType type;
 
-        this.name = name;
+    public SubscriptionPack() {
     }
 
-    public SubscriptionPack() {
+    public SubscriptionPack(String name) {
```
larubbio: Should we allow this constructor?  It seems like a type should always be provided.

larubbio:
@llavoie Other than the four comments it looks ok.

llavoie:
Danke schoen. On it.

llavoie:
@larubbio  Just pushed another commit

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
```diff
@@ -62,7 +63,7 @@ public InboxResponse getInboxDetails(@RequestParam(required = false) Long callin
             throw new NotFoundException(String.format(NOT_FOUND, "callingNumber"));
         }
 
-        Set<Subscription> subscriptions = subscriber.getSubscriptions();
+        Set<Subscription> subscriptions = subscriber.getActiveSubscriptions();
```
larubbio: If we end up doing Early Subscriptions then this will have to change.  I think it is ok for now.

larubbio:
@llavoie lgtm

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
llavoie: Yep, we have #160 for early subscriptions, so can modify when we get to it.

PR #201, opened 2015-04-30
Fixes #168 Save Deactivation Reason

### LINKED ISSUE #168: NMS.KK.ACCESS.016 - Save deactivation reason
Kilkari shall store the reason for deactivation (e.g. from user as per #161 NMS.KK.ACCESS.009 or by an event as per #116 NMS.GEN.MCTS.008.

llavoie:
Adds a DeactivationReason enum and passes it to deactivateSubscription method


koshalt:
LGTM

PR #202, opened 2015-04-30
Mobile Academy controller and service

### LINKED ISSUE #17: api/ma - Get MA Course

koshalt:
Adding controller and service for mobile academy.
Fixes #17 #18 #19 #20


api/src/main/java/org/motechproject/nms/api/web/MobileAcademyController.java
```diff
@@ -0,0 +1,131 @@
...
+     * @param callingNumber number of the caller
+     * @param callId unique tracking id for the call
+     * @return serialized json bookmark response
+     */
+    @RequestMapping(
+            value = "/bookmarkWithScore",
+            method = RequestMethod.GET)
+    public GetBookmarkResponse getBookmarkWithScore(@RequestParam Long callingNumber,
+                                                 @RequestParam Long callId) {
+
```
llavoie: Validate callingNumber and callId? (see validation methods in BaseController)

api/src/main/java/org/motechproject/nms/api/web/MobileAcademyController.java
```diff
@@ -0,0 +1,131 @@
...
+    @ResponseStatus(HttpStatus.OK)
+    public void saveBookmarkWithScore(@RequestBody SaveBookmarkRequest bookmarkRequest) {
+
+        Long callingNumber = bookmarkRequest.getCallingNumber();
+        if (callingNumber == null || callingNumber < 1000000000) {
+            throw new IllegalArgumentException("Not a valid calling number: " + callingNumber);
+        }
+        if (bookmarkRequest.getCallId() == null) {
+            throw new IllegalArgumentException("Invalid callId " + bookmarkRequest.getCallId());
+        }
```
llavoie: See the validation methods in BaseController that you can reuse for these parameter validations.

Also note that the string passed to IllegalArgumentException should be of the format "<failureReason: description>".

api/src/main/java/org/motechproject/nms/api/web/MobileAcademyController.java
koshalt: Got it. Will fix

api/src/main/java/org/motechproject/nms/api/web/MobileAcademyController.java
koshalt: Got it. will fix.

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/CourseResponse.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Response object for get course API 2.2.2
+ */
+public class CourseResponse {
+
+    // course name
+    private String name;
+
+    // course version in epoch datetime format
+    private int courseVersion;
```
llavoie: Will we be verifying/enforcing the format anywhere?

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/CourseVersionResponse.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public CourseVersionResponse(int courseVersion) {
+        this.courseVersion = courseVersion;
+    }
+
+    public CourseVersionResponse(Integer courseVersion) {
+        this.courseVersion = courseVersion;
+    }
+
```
llavoie: Are both (int/Integer) constructors needed?

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/GetBookmarkResponse.java
```diff
@@ -0,0 +1,31 @@
...
+ * Bookmark response object for bookmarkWithScore API 2.2.3.2
+ */
+
+public class GetBookmarkResponse {
+
+    // actual bookmark index in course
+    private String bookmark;
+
+    private Map<String, Integer> scoresByChapter;
+
```
llavoie: No constructors?

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/SaveBookmarkRequest.java
```diff
@@ -0,0 +1,49 @@
...
+public class SaveBookmarkRequest {
+
+    private Long callingNumber;
+
+    private Long callId;
+
+    private String bookmark;
+
+    private Map<String, Integer> scoresByChapter;
+
```
llavoie: I think it's nice to explicitly define the nullary (parameter-less) constructor, as you've done for most of the contract classes, even if the java compiler will generate one for you. Also a reminder that Jackson uses this constructor to convert the request JSON to your request object.

api/src/test/java/org/motechproject/nms/api/osgi/MobileAcademyControllerBundleIT.java
```diff
@@ -0,0 +1,80 @@
...
+                TestContext.getJettyPort());
+        SaveBookmarkRequest bookmarkRequest = new SaveBookmarkRequest();
+        bookmarkRequest.setCallId(BaseController.SMALLEST_15_DIGIT_NUMBER - 1);
+        HttpPost request = RequestBuilder.createPostRequest(endpoint, bookmarkRequest);
+        assertTrue(SimpleHttpClient.execHttpRequest(request, HttpStatus.SC_BAD_REQUEST, RequestBuilder.ADMIN_USERNAME, RequestBuilder.ADMIN_PASSWORD));
+    }
+
+    @Test
+    public void testSetValidBookmark() throws IOException, InterruptedException {
+
```
llavoie: Should this test be renamed testSetInvalidBookmark? You're expecting the request to fail, right?

llavoie: Also, some of the negative test cases for Kilkari (and perhaps the UserController, not sure) validate the format of the bad response (this is now possible after Frank's SimpleHttpClient enhancements). It would be nice to do that for these test cases too.

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/CourseVersionResponse.java
koshalt: No, oversight on my part. I was playing around with return types before our team decided on non-nullable returns for api response for required fields. Will fix.

kilkari/pom.xml
```diff
@@ -79,7 +79,9 @@
                         <Context-Path>kilkari</Context-Path>
                         <Export-Package>org.motechproject.nms.kilkari.service;version=${project.version},
                             org.motechproject.nms.kilkari.repository;version=${project.version},
-                            org.motechproject.nms.kilkari.domain;version=${project.version},</Export-Package>
+                            org.motechproject.nms.kilkari.domain;version=${project.version},
+                            org.motechproject.nms.kilkari.dto;version=${project.version},
+                        </Export-Package>
```
llavoie: I'm confused here -- I don't see a dto package anywhere under kilkari.

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/SaveBookmarkRequest.java
koshalt: Interesting. The default argument-less constructor gets auto-generated *if* no other constructor is defined, which explains why my IT for this didnt fail. I will add the constructor explicitly to keep up with the standard though.

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/GetBookmarkResponse.java
koshalt: same as the other class. Will add it.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentMenu.java
```diff
@@ -0,0 +1,44 @@
...
+@Entity
+public class ChapterContentMenu {
+
+    /**
+     * This is id for the End menu file of the chapter in the format”Chapter<ChapterId>_EndMenu”, where
+     * chapterId varies from 01 to 11.
+     */
+    @Field
+    private String identifier;
+
```
llavoie: just a nitpick, but why use "identifier" when we use "id" elsewhere?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentScore.java
```diff
@@ -0,0 +1,49 @@
...
+@Entity
+public class ChapterContentScore {
+
+    /**
+     * This is a id for the Score files of the chapter in the format ”Chapter<ChapterId>_Score”, where
+     * chapterId varies from 01 to 11.
+     */
+    @Field
+    private String identifier;
+
```
llavoie: same comment as above

kilkari/pom.xml
koshalt: Good catch. I have no idea how that snuck in. Will remove.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/LessonContentLesson.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+/**
+ * Contains details about actual content files to be played while playing a lesson.
+ */
+@Entity
+public class LessonContentLesson {
+
```
llavoie: Having classes called Lesson, LessonContent, and LessonContentLesson is confusing for me. :) Could LessonContentLesson be LessonContentDetails or something similar?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentMenu.java
koshalt: Id collides with the auto-generated id field in MDS. Hence, we will have call it something else here and then translate it back to id in our API response

larubbio: Can you call it chapterId like the comment?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/LessonContentLesson.java
koshalt: Trying to keep up with the naming/nesting structure that Aricent came up with :( . I really want to fix this too but it would be harder to map what goes where with their schema if I diverge

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public MaBookmark getBookmark(Long callingNumber, Long callId) {
+
+        List<Bookmark> bookmarks = bookmarkDataService.findBookmarksForUser(callingNumber.toString());
+        if (CollectionUtils.isEmpty(bookmarks)) {
+            return null;
+        } else {
+
+            if (bookmarks.size() > 1) {
+                LOGGER.debug("Found more than 1 instance of valid course, picking top");
```
llavoie: Should this read "Found than one instance of valid bookmark..." rather than course?

koshalt: Yup, will fix.

mobile-academy/src/test/resources/bootstrap.properties
```diff
@@ -2,5 +2,5 @@ sql.url=jdbc:mysql://localhost:3306/
 sql.user=root
 sql.password=password
 sql.driver=com.mysql.jdbc.Driver
-tenant.id=mtenant
+tenant.identifier=mtenant
 config.source=FILE
```
llavoie: why?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentMenu.java
llavoie: +1 on Rob's rec -- and then use similar naming for the other Entity classes that have "identifier" members.

koshalt: Yes. Logically, this is referring to the chapterId but siting about 2 levels deeper inside of a ChapterContentMenu object and more to do with the menu than the chapter itself. I just couldn't decide on a good name. I can change this to chapterId if that is the recommendation

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/LessonContentLesson.java
llavoie: Hmmm. Why do we need to maintain consistency with their schema at our data layer? We've changed schema for Kilkari so they're already going to have to deal with change there. Or do you mean you're trying to keep consistency with the hierarchy that they defined for the JSON/API layer?

mobile-academy/src/test/resources/bootstrap.properties
koshalt: Oops. Intellij doing a blanket refactor for me. reverted.

llavoie:
This builds and tests pass on my machine. Looking pretty good -- my comments are mostly minor issues, so don't be overwhelmed by the number of them. :)

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentMenu.java
larubbio: This is for the bookmark right?  If so what about bookmark or bookmarkId?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/LessonContentLesson.java
koshalt: We are not pushing for a spec-change to the MA response schema at this time. Given that assumption, I have the option to map their schema as-is (as close as possible) to our domain model or diverge.

I figured that it would be better to diverge and change the API spec at the same time than just 1 and have a hard time during debugging, trying to map theirs to ours. Happy to change it though..

llavoie: I don't feel too strongly about this. :) What you're doing makes sense.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/course/ChapterContentMenu.java
koshalt: This is used as a key into IMI's cms system, if I remember correctly and a mapping to their folder structure. This particular field is entirely related to their CMS and IVR menu and nothing to do with the course. 

It is kind of 'bookmarky' but not necessarily a bookmark for the course.

api/src/test/java/org/motechproject/nms/api/osgi/MobileAcademyControllerBundleIT.java
koshalt: Yes, you are right about the test name. Updated it and added a positive test as well.

And yes for the bad response validation.. but wanted to get something in before I dove into more IT and UT validations. Next checkin?

llavoie: Next checkin is fine.

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/CourseResponse.java
koshalt: Yes. I was planning on doing the validation in the next checkin where the MA service domain object gets mapped to the MA API response object.

PR #203, opened 2015-04-30
Fixes #123

### LINKED ISSUE #123: NMS.GEN.LANG.005 - Use existing languageLocationCode
Once the “Language-location code” of an FLW/MCTS-beneficiary is determined by any of the requirements listed above, the same “Language-location code” shall be used in future and the language determination process shall not be repeated.

This also applies to the case across MA and MK service i.e. “Language-location code” determined for a MSISDN number during MA service shall be applicable to MK service and vice versa.

larubbio:
Whenever we have a language for a user we should return it when the user is loaded via an API.

We needed to create a user for a subscription and assign them a language if they didn't already have one.

### DIFF
--- a/api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
+++ b/api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
@@ -18,11 +18,11 @@ import org.motechproject.nms.api.web.contract.kilkari.SubscriptionRequest;
 import org.motechproject.nms.flw.repository.FrontLineWorkerDataService;
 import org.motechproject.nms.flw.repository.ServiceUsageCapDataService;
 import org.motechproject.nms.flw.repository.ServiceUsageDataService;
-import org.motechproject.nms.flw.service.FrontLineWorkerService;
 import org.motechproject.nms.kilkari.domain.*;
 import org.motechproject.nms.kilkari.repository.SubscriberDataService;
 import org.motechproject.nms.kilkari.repository.SubscriptionDataService;
 import org.motechproject.nms.kilkari.repository.SubscriptionPackDataService;
+import org.motechproject.nms.kilkari.service.SubscriberService;
 import org.motechproject.nms.kilkari.service.SubscriptionService;
 import org.motechproject.nms.language.domain.Language;
 import org.motechproject.nms.language.repository.CircleLanguageDataService;
@@ -57,6 +57,9 @@ public class KilkariControllerBundleIT extends BasePaxIT {
     private static final String ADMIN_PASSWORD = "motech";
 
     @Inject
+    private SubscriberService subscriberService;
+
+    @Inject
     private SubscriptionService subscriptionService;
 
     @Inject
@@ -67,9 +70,6 @@ public class KilkariControllerBundleIT extends BasePaxIT {
 
     @Inject
     private SubscriptionDataService subscriptionDataService;
-
-    @Inject
-    private FrontLineWorkerService frontLineWorkerService;
 
     @Inject
     private FrontLineWorkerDataService frontLineWorkerDataService;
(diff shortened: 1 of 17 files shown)


api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -101,27 +105,29 @@ private void cleanAllData() {
         languageDataService.deleteAll();
     }
 
+    private void createLanguageAndSubscriptionPacks() {
+        gLanguage = languageDataService.create(new Language("tamil", "10"));
```
koshalt: nice!

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
```diff
@@ -272,6 +279,25 @@ public void testCreateSubscriptionRequestDifferentPacks() throws IOException, In
     }
 
     @Test
+    public void testCreateSubscriptionsNoLanguageInDB() throws IOException, InterruptedException {
+        setupData();
```
koshalt: We could probably take this call to setupData() out of here and just decore the setupData() method with @before to be run before every test

frankhuster: Except it's not called in the @Test method above.

koshalt: Line 238? :)

On a bigger note though, @before is probably a good standard to follow

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -39,12 +36,10 @@
     @Field
```
koshalt: Unrelated to the current change, but why use LocalDate vs DateTime in Utc?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public Subscriber getSubscriber(long callingNumber) {
+        return subscriberDataService.findByCallingNumber(callingNumber);
+    }
+
+    @Override
+    public void add(Subscriber subscriber) {
+        subscriberDataService.create(subscriber);
```
koshalt: Are we keeping the logic to check for existing subscriber outside of this class?

koshalt:
minor non-blocking comments. lgtm for a merge

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
frankhuster: Damn it! It was hidden in the diff...
loser

frankhuster: It only makes sense to have a @before if you need the database to be in the exact same state for each and every test, which is not always the case.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
frankhuster: I was wondering the same thing. We should be consistent system-wide .

llavoie: What are we doing elsewhere? Agree that we should be consistent.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
llavoie: We should do it here (although the API layer also does check, I believe). I can add in my next checkin.

llavoie: Never mind, I just realized that this is a new method added by Frank, not an existing one. @frankhuster should probably fix this then. :)

llavoie: Err, I mean @larubbio. Sigh. Didn't realize which pull request I was looking at.

api/src/test/java/org/motechproject/nms/api/osgi/KilkariControllerBundleIT.java
koshalt: We could be making the same point here but every test (except for 3, which would be affected anyways) call setupData here, which in turn drops the data and creates new stuff. I'm just saying that it is good practice to refactor common stuff into a 'before' if the majority of the tests need to do the same thing.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
larubbio: It isn't really a new method, I just moved it out of SubscriptionService.  

Although the Subscriber table is unique in the database so that would throw if this was a dupe.  Is that the check you would want to do in code as well?

PR #207, opened 2015-05-01
fileAuditing, better targetFile IT

frankhuster:


outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/domain/AuditRecord.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+import org.motechproject.mds.annotations.UIDisplayable;
+
+@Entity(tableName = "nms_obd_audit_records")
+public class AuditRecord {
+    @Field
+    @UIDisplayable(position = 0)
+    private FileType type;
```
larubbio: Are any of these fields allowed to be null in the db?  If not should you add @Column(allowsNull = "false")

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/domain/FileType.java
```diff
@@ -0,0 +1,6 @@
+package org.motechproject.nms.outbounddialer.domain;
+
+public enum FileType {
+    CdrFile,
```
larubbio: Should these be all caps?

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/domain/AuditRecord.java
frankhuster: ok will do

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
```diff
@@ -113,13 +117,15 @@ public TargetFileServiceImpl(@Qualifier("outboundDialerSettings") SettingsFacade
                                  MotechSchedulerService schedulerService, AlertService alertService,
                                  SubscriptionDataService subscriptionDataService,
                                  CallRetryDataService callRetryDataService,
-                                 SubscriberDataService subscriberDataService) {
+                                 SubscriberDataService subscriberDataService,
```
larubbio: Should this use the subscriberDataService, or the subscriberService?  We've never come up with clear rules for when to use which.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/domain/FileType.java
frankhuster: yep

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
```diff
@@ -130,6 +136,16 @@ private String targetFileName() {
...
+        } else {
+            fileAuditDataService.create(new AuditRecord(FileType.TargetFile, tfn.getFileName(), tfn.getRecordCount(),
+                    tfn.getChecksum(), status));
+        }
+    }
+
+
     private File createTargetFileDirectory() {
         File userHome = new File(System.getProperty("user.home"));
         File targetFileDirectory = new File(userHome, settingsFacade.getProperty(TARGET_FILE_DIRECTORY));
```
larubbio: I think they might be giving us an ftp location to place the file.  I think the File class will handle that but I'm not sure.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
```diff
@@ -130,6 +136,16 @@ private String targetFileName() {
     }
 
 
+    private void insertTargetFileAuditRecord(TargetFileNotification tfn, String status) {
```
larubbio: Instead of the if check should you just have an overloaded method?

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
frankhuster: We should be able to symlink a network share in $HOME, right?

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
frankhuster: Right.
Should we create a "harmonize all cross-module database access" ticket?

outbound-dialer/src/test/java/org/motechproject/nms/outbounddialer/it/TargetFileServiceBundleIT.java
```diff
@@ -102,20 +111,33 @@ private void setupDatabase() {
...
+        //read the file to get checksum & record count
+        File homeDir = new File(System.getProperty("user.home"));
+        File targetDir = new File(homeDir,
+                settingsService.getSettingsFacade().getProperty("outbound-dialer.target_file_directory"));
+        File targetFile = new File(targetDir, tfn.getFileName());
+        MessageDigest md = MessageDigest.getInstance("MD5");
+        int recordCount = 0;
+        try (InputStream is = Files.newInputStream(targetFile.toPath());
+             BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
+            DigestInputStream dis = new DigestInputStream(is, md);
```
larubbio: Is this correct?  I see you instantiate dis but it doesn't look like it is used.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
larubbio: Probably.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
larubbio: a network share yes, but not an ftp server.  Also I worry about maintaining that mount across all servers in the cluster.

I don't have the SRS open, but we do need to know if it is a network share or ftp.  (I don't know, you might know better than me).  If not just mark it with a todo and a ticket.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
frankhuster: How about I remove the if altogether and pass `new TargetFileNotification()` instead of `null`?

larubbio:
In general lgtm.  Also the babysitter is about to arrive so I'm outta here.....

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
larubbio: That would work.  You'd have to check that tfn wasn't null.

outbound-dialer/src/test/java/org/motechproject/nms/outbounddialer/it/TargetFileServiceBundleIT.java
frankhuster: That's the way you use these, it's a sort of listener that silently builds the checksum as you go about the business of reading or writing the file.

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
frankhuster: #209

Issue #99, opened 2015-04-25
NMS.GEN.LOC.003 - Location Validation Checks during upload


larubbio:
The NMS system shall perform the following checks and validation before uploading location data into the location database:
 - None of the Mandatory Location fields can be blank
 - Duplicate location data does not already exist in the system

The exact definitions of these fields are provided in Appendix A (refer section 6.2).

===== Non SRS comment below ======
When this is finalized the SRS should be updated.  I've created this google sheet to track the format:

https://docs.google.com/spreadsheets/d/1rIjiou_DuJ1BgpXqImsecOK7jbAFFLOfZKAYgiCQQto/edit?usp=sharing

larubbio:
Currently I don't have any spec that says which fields are mandatory and which fields determine uniqueness.   I asked Pankaj if he could track it down.

The current code assumes the code is unique.  It also validates that each child has a parent.  i.e. a district must have a state.  It does not validate that parents have children (so a state can have 0 districts)

larubbio:
From:
https://drive.google.com/open?id=0BzVwhkviOT8dekNIbk1LX3ZERnc&authuser=1

![screen shot 2015-04-29 at 9 41 41 am](https://cloud.githubusercontent.com/assets/51527/7396211/339a0586-ee54-11e4-92e9-cfc75ffec9fe.png)

It is still unclear if this is the complete location file, and what the individual fields store.

larubbio:
From a file called master.xlsx:
![image](https://cloud.githubusercontent.com/assets/51527/7424435/d1fca20e-ef55-11e4-9341-b85499d5b316.png)

That file can be imported to build out the location database (although the village name is unclear.  Why do some have an '*' and others have numbers in parens?)

The relationship between that master file and the  "Health_NonCensus_Village_master" file is not clear since the master file has village information in it.  

Will the location data come in the form of the single master file or will there be three files, the master, Health_NonCensus_Village_Master.docx and Census_vllage_master.docx files that together make up the full location data?

Will there always be all levels of the hierarchy?  I think the only optional level is Sub center, but I'm not sure.

larubbio:
If SVID is the NON Census village id and VCode is the village code will the both be present for all villages, or will SVID be set only for non-census villages and will VCode be set only for census villages?

Assuming a census village, will it have just VCode, just SVID or both?
Assuming a non-census village, will it have just SVID, just VCode or both?

larubbio:
[...]

PR #208, opened 2015-05-01
Interim commit

larubbio:
This adds database constraints and some fields that I think are required.  It is not the complete fix for #99.  I'm still waiting for final word on the location requirements.  I'm also waiting for a fix to the detached field issue.  I would like to merge this so the work doesn't get stale.

### DIFF
--- a/location/src/main/java/org/motechproject/nms/location/domain/HealthBlock.java
+++ b/location/src/main/java/org/motechproject/nms/location/domain/HealthBlock.java
@@ -3,9 +3,9 @@ package org.motechproject.nms.location.domain;
 import org.motechproject.mds.annotations.Cascade;
 import org.motechproject.mds.annotations.Entity;
 import org.motechproject.mds.annotations.Field;
-import org.motechproject.mds.annotations.UIDisplayable;
 import org.motechproject.mds.domain.MdsEntity;
 
+import javax.jdo.annotations.Column;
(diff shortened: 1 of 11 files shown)


location/src/main/java/org/motechproject/nms/location/domain/HealthBlock.java
```diff
@@ -19,18 +19,18 @@
 public class HealthBlock extends MdsEntity {
 
     @Field
-    @UIDisplayable(position = 0)
+    @Column(allowsNull = "false")
```
frankhuster: What's the difference between `@Column(allowsNull = "false")` and `@NotNull`?

frankhuster:
Oh, and why remove the UI ordering?

frankhuster:
lgtm

location/src/main/java/org/motechproject/nms/location/domain/HealthBlock.java
larubbio: @Column(allowsNull = "false") adds a not null constraint to the database.  @NotNull does a validation in code and throws an exception before saving if the value is null.  So one is in code and the other is in the database.  (If I had to choose only one I would go with @Column

larubbio:
I figured we could add the UI ordering if there is a story/need for it.

Issue #180, opened 2015-04-26
NMS.KK.DND.001 - DND database


larubbio:
NMS shall not initiate any OBD call to a beneficiary if the beneficiary has the MSISDN registered in the DND database and the activation has happened via MCTS import.

The DND restriction does not apply for those beneficiary that have activated via the IVR.

To realize this requirement, NMS system shall maintain a DND database. The exact mechanism to realize the DND database is outside the scope of this document.

The DND restrictions apply to OBD call for Kilkari service only. They do not apply to incoming services like MA and MK. For MA, it also does not apply to course-completion SMS.

larubbio:
There may not be work for this item.  I've emailed IMI to ask about their DND handling.

larubbio:
From the email thread:

    The handling for following requirement is missing in API document. To fulfill this requirement, 
    a new field is needed in OBD record to convey the source of kilkari subscription i.e. IVR/MCTS. 
    Please let me know, if this change is fine with you.

llavoie:
My current in-progress change adds a "subscription mode" to the subscription, with values IVR and MCTS_UPLOAD. The mode is passed at subscription time (the REST API always passes IVR) and certain behavior is determined by it (e.g. rules for rejecting subscriptions detailed in #159).

llavoie:
@frankhuster  You'll want to add one more column to the OBD target file you are generating, which contains the Subscriber.subscriptionMode (IVR or MCTS).

PR #211, opened 2015-05-04
Complete file format for targetFile

frankhuster:
and also dismisses magic number PMD errors in annotations (ie: mostly column widths)

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -163,6 +163,7 @@
             <property name="severity" value="info"/>
             <property name="tokens" value="NUM_DOUBLE,NUM_FLOAT,NUM_INT,NUM_LONG"/>
             <property name="ignoreNumbers" value="-1,0,1,2,60,24,7,31,32"/> <!-- 60,24 & 7 are used when calculating dates.  31 in hash. 32 is Magic Johnson's number so it gets a free pass -->
+            <property name="ignoreAnnotation" value="true" />
         </module>
         <module name="IllegalInstantiation">
             <property name="severity" value="warning"/>
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
@@ -81,9 +81,9 @@ public class Subscription {
 
     /**
      * Returns the weekId corresponding to today's date.
-     * @return
      */
-    @Ignore int todaysWeekId() {
+    @Ignore
+    int todaysWeekId() {
         //todo: the work
         return 0;
     }
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
@@ -3,6 +3,7 @@ package org.motechproject.nms.kilkari.domain;
 import org.joda.time.LocalDate;
 import org.motechproject.mds.annotations.Entity;
 import org.motechproject.mds.annotations.Field;
+import org.motechproject.mds.annotations.Ignore;
 
 import javax.jdo.annotations.Column;
 import javax.jdo.annotations.Unique;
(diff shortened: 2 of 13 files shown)


checkstyle.xml
```diff
@@ -163,6 +163,7 @@
             <property name="severity" value="info"/>
             <property name="tokens" value="NUM_DOUBLE,NUM_FLOAT,NUM_INT,NUM_LONG"/>
             <property name="ignoreNumbers" value="-1,0,1,2,60,24,7,31,32"/> <!-- 60,24 & 7 are used when calculating dates.  31 in hash. 32 is Magic Johnson's number so it gets a free pass -->
+            <property name="ignoreAnnotation" value="true" />
```
larubbio: Will I no longer have to do this:

```java
   // checkstyle made me do it
    public static final String FALSE_STRING = "false";

    @Field
    @Column(allowsNull = FALSE_STRING, length = 100)
    @NotNull
    @Size(min = 1, max = 100)
    private String name;
```

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/service/impl/TargetFileServiceImpl.java
```diff
@@ -165,27 +168,203 @@ private File createTargetFileDirectory() {
     }
 
 
-    private void writeSubscriptionRow(Subscription subscription, OutputStreamWriter writer) throws IOException {
-        //subscription id
-        writer.write(subscription.getSubscriptionId());
+    private void writeSubscriptionRow(String fileIdentifier, Subscription subscription, String imiServiceId,
```
llavoie: Per #180, we need to add one more column to the file that tells IMI how the user was subscribed (IVR or MCTS_UPLOAD), which you can get from the Subscription.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -80,6 +79,15 @@ public void setDeactivationReason(DeactivationReason deactivationReason) {
         this.deactivationReason = deactivationReason;
     }
 
+    /**
+     * Returns the weekId corresponding to today's date.
+     * @return
+     */
+    @Ignore int todaysWeekId() {
```
koshalt: how does the @Ignore work here?

outbound-dialer/src/main/java/org/motechproject/nms/outbounddialer/domain/AuditRecord.java
```diff
@@ -8,35 +8,62 @@
...
 @Entity(tableName = "nms_obd_audit_records")
 public class AuditRecord {
+    /**
+     * The identifier field is used in each targetFile csv row combined with the subscription id to form a uniquely
+     * identifying id for a specific OBD record
+     */
     @Field
     @UIDisplayable(position = 0)
-    private FileType type;
+    private String identifier;
```
koshalt: Not sure if I understand the comments above. The identifier here is then combined with the subscription id to make the unique id? is this the batch id or request id that IMI wants?

koshalt: I think I understand it now, looking at the code further down. Can we call this something else like fileId or batchId? Mostly a nitpick though

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
larubbio: It tells MDS to not create a database field for that member.

koshalt:
mostly minor comments. lgtm otherwise. I'll let you do the merge after you've had a chance to look at the comments

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
koshalt: Ok, makes sense. Any reason why we wouldn't have an access modifier on the field? (public/private)

frankhuster: It's not a field it's a method!

PR #217, opened 2015-05-05
Fixes #120

### LINKED ISSUE #120: NMS.GEN.LANG.002 - circle to state/district
NMS shall maintain a mapping between a telecom circle and the associated districts of a state contained in a circle. 

This mapping shall be used to determine the state and consequently the “Language-location code” used in the state.

A telecom circle can have various mapping with states:
- One-circle-to-one-state mapping (e.g. AP)
- One-circle-to-many-state mapping (e.g. North-east, Bihar + Jharkhand or MP + Chattisgarh)
- Many-circle-to-one-state mapping (e.g. UP-east and UP-west)

larubbio:
This commit finishes the merging of location and language and gets the tests to pass.  It fixes #120 I'm still concerned about the way many-to-many relationships are handled.  I think the api is error prone.


region/src/main/java/org/motechproject/nms/region/location/service/LocationService.java
```diff
@@ -0,0 +1,11 @@
+package org.motechproject.nms.region.location.service;
+
+import org.motechproject.nms.region.language.domain.Language;
+import org.motechproject.nms.region.location.domain.State;
+
+/**
+ * Service interfaces exposed by the language module
```
koshalt: region or location module?

koshalt:
lgtm.

PR #219, opened 2015-05-06
Update module references that were missed in original commit

larubbio:


llavoie:
Merging even though CI fails because our CI appears to be confused right now.

PR #226, opened 2015-05-08
Remove col size constants

frankhuster:
Since `checkstyle.xml` is now changed to allow magic numbers in annotations I replaced ugly:

```java
@Column(length = FIELD_SIZE_10)
```

with prettier:

```java
@Column(length = 10)
```


frankhuster:
Aargh, this picked up my previous & unrelated changes...
I'll back this out and add it later.
:imp:

Issue #216, opened 2015-05-05
NMS.KK.CDR.001 Call Details Record processing


frankhuster:
NMS API - 4.2.6.1 cdrFileNotification API - Request
CDR summary & CDR detail described in NMS API - 4.1.5 OutBound Dialer Service bullet point 14

PR #228, opened 2015-05-09
Now populating the CallRetry table

frankhuster:
The latest commit now has functional (but not exhaustive) ITs
Partial fix for #216 

[LL: updating your description so that this doesn't auto-close a partially fixed ticket]


imi/src/main/java/org/motechproject/nms/imi/service/impl/RequestId.java
```diff
@@ -0,0 +1,43 @@
...
+        return fileIdentifier;
+    }
+
+    public String getSubscriptionId() {
+        return subscriptionId;
+    }
+
+    public static final RequestId fromString(String s) {
+        String fileIdentifier = s.substring(0, UUID_LENGTH);
+        String subscriptionId = s.substring(UUID_LENGTH, UUID_LENGTH);
```
koshalt: is this supposed to be s.substring(16, 16) or (16,32) ?

frankhuster: Oops, I guess the fact that I haven't really written any tests for that does show, eh..

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -198,4 +199,21 @@ public SubscriptionPack getSubscriptionPack(String name) {
         return subscriptionPackDataService.byName(name);
     }
 
+
+    /**
+     * To be used by ITs only!
+     */
+    public void deleteAll() {
```
koshalt: didn't realize that we were going to force ITs to use the service too (instead of data layer)...but it makes sense

koshalt:
lgtm, the uuid comment in RequestId is the only item that needs attention. ok to fix and merge.

imi/src/main/java/org/motechproject/nms/imi/service/RequestId.java
```diff
@@ -0,0 +1,43 @@
...
+    public String getFileIdentifier() {
+        return fileIdentifier;
+    }
+
+    public String getSubscriptionId() {
+        return subscriptionId;
+    }
+
+    public static final RequestId fromString(String s) {
+        String fileIdentifier = s.substring(0, UUID_LENGTH);
```
koshalt: do you need to do any validation here to make sure that the request id is in the expected format?

imi/src/main/java/org/motechproject/nms/imi/service/impl/ReschedulerServiceImpl.java
```diff
@@ -19,11 +26,16 @@
...
 
 
     @Autowired
-    public ReschedulerServiceImpl(EventRelay eventRelay) {
+    public ReschedulerServiceImpl(EventRelay eventRelay, SubscriptionService subscriptionService,
+                                  CallRetryDataService callRetryDataService) {
         this.eventRelay = eventRelay;
+        this.subscriptionService = subscriptionService;
+        this.callRetryDataService = callRetryDataService;
     }
```
koshalt: How do you feel about moving this part to the caller? So, CdrFileServiceImpl just adds the data to an event and publishes it with a subject. That way, this end up being being a pure service class that know how to listen to specific events and then process the data.

I understand the need to raise events and distribute load but It feels a little weird that the method exists to raise an event that is then captured by another method in the same class. this is borderline nitpick though.

imi/src/main/java/org/motechproject/nms/imi/service/RequestId.java
frankhuster: I don't think so.

Issue #119, opened 2015-04-25
NMS.GEN.LANG.001 - State/District to languageLocationCode mapping


larubbio:
NMS shall maintain a mapping between a geographical location of a state (i.e. set of districts) and the “Language-location code” used in that location. 

There shall be one and only one “Language-location code” used in a given geographical location.
This “Language-location code” shall be same and applicable to all services in that location. 

The mapping shall be uploaded in NMS when one or more services are deployed in a state.

If there is only one “Language-location code” in the state, all districts shall have same “Language-location code”; else, multiple groups shall be formed, each group having a set of districts and having a unique “Language-location code”.

Division or reorganization of geographical location for “Language-location code” mapping is not supported and an item for future.

larubbio:
I opened #218 to cover:

> There shall be one and only one “Language-location code” used in a given geographical location.
> This “Language-location code” shall be same and applicable to all services in that location.

larubbio:
![screen shot 2015-05-06 at 11 11 01 am](https://cloud.githubusercontent.com/assets/51527/7500108/a1c3c830-f3e0-11e4-98b3-8f39bc7aa484.png)

larubbio:
This issue should have closed when #119 was merged into master.

PR #230, opened 2015-05-09
#119

larubbio:
Add LanguageLocation class and combine language, location and circle modules into the region module


frankhuster:
test this please

larubbio:
it will fail. :(

frankhuster:
I wanted to see what the results of the build were, the prior build had been erased...

larubbio:
This is a live look in at the server room:

frankhuster:
eeek

larubbio:
I think this should build and pass now.  I had to do a ton of merging from @frankhuster code.  I then modified the datamodel to work around the cascade issue.  I then got everything working and was about to submit but .... I had to merge more of @frankhuster dang code.

Once the cascade issue is fixed the datamodel should be reverted.  Ideally instead of LangaugeLocation having a boolean defaultForCircle field, the Circle would point to it's default LanguageLocation.  This way having only one default is enforced in the datamodel.

cc @llavoie @koshalt @frankhuster

region/src/main/java/org/motechproject/nms/region/domain/validation/VillageValidator.java
```diff
@@ -15,10 +15,17 @@ public void initialize(ValidVillage validVillage) {
 
     @Override
     public boolean isValid(Village village, ConstraintValidatorContext constraintValidatorContext) {
+        if (village == null) {
+            return true;
+        }
+
         if (village.getVcode() == null && village.getSvid() == null) {
+            constraintValidatorContext
```
frankhuster: Sweet!

larubbio: Except it currently throws an npe.  I put a ticket in platform for it.

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Autowired
+    public LanguageLocationServiceImpl(LanguageLocationDataService languageLocationDataService) {
+        this.languageLocationDataService = languageLocationDataService;
+    }
+
+    /**
+     * Given a languageLocationCode returns the object if one exists
+     * @param code the code for the language location record
+     * @return The languageLocationRecord if it exists
```
frankhuster: LanguageLocation object

frankhuster:
So as of now, we're not recording history for Circle & all?
If we're cool with that, then lgtm.
Otherwise, we can fix in a subsequent PR.

koshalt:
Should we create a tracking item to audit history tracking in the next
milestone?

larubbio:
There was another issue with platform around history tracking (the stack
overflow).  I put an issue in for it.

larubbio:
I reopened the history tracking issues and moved them to sprint 3

PR #235, opened 2015-05-12
Fixes #216, fixes #170

### LINKED ISSUE #170: NMS.KK.DND.002 - Rules for deactivation related to DND (Do Not Disturb)
NMS shall mark a beneficiary as deactivated with associated reason if the beneficiary has its MSISDN number added to the DND database. This requirement applies to both cases listed below:
- The beneficiary is called for first time 
- The beneficiary has added its MSISDN number to the DND database while the OBD service is ongoing

In case a beneficiary removes its number from DND database, then such a use-case is currently not handled by the software.

### LINKED ISSUE #216: NMS.KK.CDR.001 Call Details Record processing

frankhuster:
Still some //todo comments to followup with
Changed ugly `LocalDate` references to pretty `DateTime`


imi/src/main/java/org/motechproject/nms/imi/component/CallRescheduler.java
```diff
@@ -0,0 +1,97 @@
...
+
+                //update the callStage
+                callRetry.setCallStage((callRetry.getCallStage() == CallStage.RETRY_1) ? CallStage.RETRY_2 :
+                        CallStage.RETRY_LAST);
+                //increment the day of the week
+                callRetry.setDayOfTheWeek(callRetry.getDayOfTheWeek().nextDay());
+                //update the CallRetry record
+                callRetryDataService.update(callRetry);
+            }
+        } catch (Exception e) {
```
koshalt: This feels a little risky right? Catching and swallowing a general exception? Ok to swallow known exceptions but we should crash on unknowns

frankhuster: Funny you should mention.
This is debug code I forgot to remove.
But, interestingly enough, it's a bit tricky.
Exceptions are 'eaten' when thrown inside event listeners, and the event listener is called again 3 times I think.

I'll remove it anyway.

Good eye!

imi/src/main/java/org/motechproject/nms/imi/component/SubscriptionDeactivator.java
```diff
@@ -0,0 +1,74 @@
...
+                                   AlertService alertService) {
+        this.subscriptionService = subscriptionService;
+        this.callRetryDataService = callRetryDataService;
+        this.alertService = alertService;
+    }
+
+
+    @MotechListener(subjects = { DEACTIVATE_SUBSCRIPTION })
+    public void deactivateSubscription(MotechEvent event) {
+        LOGGER.info("deactivateSubscription() is handling {}", event.toString());
```
koshalt: we should probably sit down at some point and figure out the right log levels across the services. I would think that event.toString is pretty expensive and might leak PII data? Would this be debug?

koshalt:
looks good. couple of minor comments.

oh and the build... :)
fmc

imi/src/main/java/org/motechproject/nms/imi/component/SubscriptionDeactivator.java
frankhuster: Yep, good point, should be debug. Will change than in the next commit.

PR #236, opened 2015-05-12
KK CDR processing - IT now tests for invalid rejections

frankhuster:


imi/src/test/java/org/motechproject/nms/imi/it/CdrFileServiceBundleIT.java
```diff
@@ -91,13 +99,13 @@ public void testValidRequest() throws IOException, NoSuchAlgorithmException {
...
-                        new FileInfo(helper.cdrSummaryFileName(), helper.summaryFileChecksum(), 0),
+                        new FileInfo(helper.cdrSummaryFileName(), helper.summaryFileChecksum(), 6),
                         new FileInfo(helper.cdrDetailFileName(), helper.detailFileChecksum(), 1))
         );
 
         try {
             long sleepyTime = 10 * 1000L;
-            getLogger().debug("Sleeping {} seconds to give a chance to @MotechListeners to catch up...", sleepyTime/1000);
+            getLogger().debug("Sleeping {} seconds to give a chance to @MotechListeners to catch up...", sleepyTime / 1000);
             Thread.sleep(sleepyTime);
```
llavoie: This seems a bit fragile... is there a programmatic way to check that the listeners executed rather than sleeping an arbitrary amount of time?

llavoie: Or should I not be reviewing this test yet since it has an Ignore annotation on it?

frankhuster: Ignore for now, and I'll look into making that more solid.

llavoie:
lgtm

Issue #221, opened 2015-05-06
OBD needs to indicate when a welcome message should be played rather than the first message in the pack


llavoie:
Breaking this out into a separate issue from #157, which deals mostly with setting subscription start dates.

Handling for following requirement is missing in API specification. 

"If a woman is subscribed to Kilkari via MCTS data, the service could start at any point during the 72
week or 48 week schedule. Since we cannot assume that her ASHA or ANM will have informed her
about Kilkari, we will need to play her an initial welcome message before any other content is played."

To handle this requirement a new field is needed in OBD record (in OBD file) to indicate the first 
OBD message of subscribed pack.

llavoie:
The pull request referenced above adds a "needsWelcomeMessage" field to Subscription, and sets it to true for a new MCTS subscription. The CDR processor will need to set this field to false once the subscriber has received the welcome message. Assigning to @frankhuster.

frankhuster:
Fixed by #246

PR #240, opened 2015-05-13
Fixes #163 -- adds helper methods for OBD to schedule the appropriate message to play next for subscriber

### LINKED ISSUE #163: NMS.KK.ACCESS.011 - Scheduling of weekly messages
Kilkari shall schedule the delivery of weekly messages to the beneficiary based on the activation date of the pack. Week is computed from LMP or DOB.

Also see comment in #157 NMS.KK.ACCESS.004.

llavoie:
Also includes a partial fix for #221 -- @frankhuster will handle the other part of it in the CDR processor.

Does some renaming/commenting left over from last checkin, and adds column constraints for Subscription entity.


kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -79,15 +97,74 @@ public void setDeactivationReason(DeactivationReason deactivationReason) {
...
-     * Returns the weekId corresponding to today's date.
+     * Helper method to be called by the OBD process when selecting a message to play for a subscription
+     * @param date The date on which the message will be played
+     * @return SubscriptionPackMessage with the details of the message to play
      */
-    @Ignore
-    int todaysWeekId() {
-        //todo: the work
-        return 0;
+    public SubscriptionPackMessage nextScheduledMessage(DateTime date) { //NO CHECKSTYLE CyclomaticComplexity
```
llavoie: I know that turning off the checkstyle CyclomaticComplexity check is kind of gross... want to get Frank unblocked, can look at simplifying this later if we care.

frankhuster:
lgtm, I'll see if there's something I can do about that cyclomatic stuff if I have time later

PR #243, opened 2015-05-13
Fixes #166

### LINKED ISSUE #166: NMS.KK.ACCESS.014 - Rule for marking subscription completed
Kilkari service shall be completed after the Service Pack runs for its scheduled duration (subscription state = Completed).

frankhuster:
The status of a subscription will change to completed during the CDR
processing when the last message has been successfully played (or not)

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/component/SubscriptionCompleter.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/component/SubscriptionCompleter.java
@@ -1,0 +1,76 @@
+package org.motechproject.nms.imi.component;
+
+import org.joda.time.DateTime;
+import org.motechproject.event.MotechEvent;
+import org.motechproject.event.listener.annotations.MotechListener;
+import org.motechproject.nms.imi.domain.CallDetailRecord;
+import org.motechproject.nms.imi.domain.CallRetry;
+import org.motechproject.nms.imi.repository.CallRetryDataService;
+import org.motechproject.nms.imi.service.RequestId;
+import org.motechproject.nms.kilkari.domain.Subscription;
+import org.motechproject.nms.kilkari.service.SubscriptionService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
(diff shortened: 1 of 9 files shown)


imi/src/main/java/org/motechproject/nms/imi/component/SubscriptionCompleter.java
```diff
@@ -0,0 +1,76 @@
...
+    @Autowired
+    public SubscriptionCompleter(SubscriptionService subscriptionService,
+                                 CallRetryDataService callRetryDataService) {
+        this.subscriptionService = subscriptionService;
+        this.callRetryDataService = callRetryDataService;
+    }
+
+
+    @MotechListener(subjects = { COMPLETE_SUBSCRIPTION })
+    public void completeSubscription(MotechEvent event) {
```
larubbio: Should this live in the IMI module or the kilkari module?

frankhuster: The IMI module I would think.
The CDR class is 'specific' to what we get back from IMI.

larubbio: Sure but don't you put it into an event and then run business logic around
the closing of subscriptions off of it?

frankhuster: Yep.
But that would make Kilkari dependent of IMI.
I guess we could create a Kilkari specific DTO instead.
Should we have the Kilkari module handle all three CDR procesing events, actually?

larubbio: I think so.  I imagine we would define a dto that the IVR interface has to produce and publish.  Then all event handling that result in changes to subscriptions would live in the kilkari module.  If we change ivr providers the new module just needs to recreate and publish those events.

frankhuster: Yep, sounds like a good idea.
Wanna merge this PR and I'll put the IMI => KK refactor in a new PR?

frankhuster:
Hey @larubbio, waiting on your answer for this guy.

larubbio:
I haven't had a chance to look at this beyond that first event listener yet
(I was doing it in between compiles last night and didn't get that far on
it).

PR #246, opened 2015-05-14
Migrated CDR to Kilkari module, IT rework not using Thread.sleep()

frankhuster:
Quite a big change. The code is now much clearer. There is still quite a bit of IT work to do.


imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
```diff
@@ -2,62 +2,39 @@
...
-    private String file;
+    @Column(allowsNull = "false")
+    private String fileName;
 
     @Field
-    @UIDisplayable(position = 3)
     @Column(allowsNull = "false")
     private String status;
 
     @Field
```
koshalt: Can recordCount be null?

imi/src/main/java/org/motechproject/nms/imi/exception/NotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package org.motechproject.nms.imi.exception;
+
+public class NotFoundException extends IllegalStateException {
+    public NotFoundException(String message) {
+        super(message);
```
koshalt: Just brainstorming here. If this is being used as a wrapper to throw 404 instead of 400 for a file request, I wonder if 404 is the right response for it.

Ideally, 404 would be a response to a GET method where we say we can't find a resource that they are trying to get... but if they are POSTing to us to go find files that don't exist, it should be considered a bad request, right?

imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
frankhuster: Yep, if there's a file-wide error (ie: file not found)

imi/src/main/java/org/motechproject/nms/imi/exception/NotFoundException.java
frankhuster: I thought we'd decided things not found in the database would always return 404

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -47,47 +41,50 @@
...
+    private void reportErrorAndThrow(String file, String error) {
+        LOGGER.error(error);
+        alertService.create(file, CDR_SUMMARY, error, AlertType.CRITICAL, AlertStatus.NEW, 0, null);
+        fileAuditRecordDataService.create(new FileAuditRecord(FileType.CDR_SUMMARY_FILE, file, error, null, null));
+        throw new IllegalStateException(error);
+    }
+
+
+
+    private CdrParseResult parseSummaryCdrs(String dir, String fileName, String checksum) {
```
koshalt: I'm a little worried (or paranoid) that this method might blow up the memory on the machine while processing a few hundred thousand records from the file. The list object will most likely end up in the large object heap allocation and we might run into problems with the GC and memory fragmentation.

Assuming I understand the design correctly, we take a file from a full day of cdrs and load it up in a memory list object. Then, we iterate through the list and fire event to process the CDRs. Would it be more efficient to fire the event as you go about parsing the file?

frankhuster: We should be able to handle multiple million CDRs in memory (will try in Sprint3)
If we get IMI to only deliver one CDR file then we could do that, but not sure how we could otherwise...

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
```diff
@@ -106,7 +118,7 @@ public void setNeedsWelcomeMessage(boolean needsWelcomeMessage) {
     }
 
     /**
-     * Helper method to be called by the OBD process when selecting a message to play for a subscription
+     * Helper method to be called by the OBD processCallDetailRecord when selecting a message to play for a subscription
```
koshalt: Unrelated to your current change, but this seems like a lot of business logic to exist in a domain object that we would consider to be a data container, right? Should be move this to a helper method? 

I'm torn on this this design pattern http://java.dzone.com/articles/business-logic-domain-objects

frankhuster: @llavoie & @larubbio, what do you think?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
koshalt: I guess I'm not worried the ability to handle a million records in memory. The comment was about the aftermath on the heap after the large list has been processed. Not sure how many files IMI is planning on giving us at a time but the concern is only real if any single file has enough records to make the list end up on the heap http://www.oracle.com/webfolder/technetwork/tutorials/obe/java/gc01/index.html

We had similar problem in one of the MS data aggregation services that I worked on. Anyways, you can file this under premature optimization and we can deal with this if it really pops up later.

koshalt:
no blocking comments. lgtm for merge.

frankhuster:
Fixes #221

PR #248, opened 2015-05-14
Fixes #102, fixes #103 -- FLW status and identification

### LINKED ISSUE #102: NMS.GEN.FLW.002 - FLW Identification
NMS shall internally generate and maintain an FLW-ID for all FLWs that is called the NMS FLW-ID.
For external interfaces (including IVR and CSV upload), the FLW shall be identified first by the NMS FLW-ID if present.  Next the MCTS FLW-ID should identify them if it is provided.  If neither of those IDs is provided then the MSISDN should be used.

3.1.2.1.1	Additionally, for FLW records coming via CSV file, the record may optionally contain an FLW-ID provided by MoHFW called the MCTS FLW-ID 

MCTS FLW-ID is optional because data uploaded using training data may not have the FLW-ID issued by MCTS.

### LINKED ISSUE #103: NMS.GEN.FLW.003 - FLW Status
The NMS system shall maintain the status of an FLW:
- **Anonymous**: Refers to any user who has called the MA/MK service and whose MSISDN number, FLW ID, name, and location data is not available in the FLW contact list maintained by the NMS system (in other words – only the MSISDN exists in the database). If an FLW record is uploaded for such a user, existing information (e.g. call records, language information) of the anonymous user shall not be deleted or over-written – however the FLW information not available in the NMS shall be added to the record (at which point they will become an 
Active FLW).
- **Active FLW**: Mandatory fields in the record of an FLW are present and following conditions are met:
  - FLW must have called NMS system once (via MA or MK)
  - FLW has a valid location  (State and District fields are mandatory and valid; other location fields if present are valid).
  - FLW has all the following parameters present in its record:
    - MSISDN number
    - Name
    - Designation
    - Mandatory parts of Location Data (NMS.GEN.LOC.002)
- **Inactive FLW**: A record of an FLW where all the mandatory fields have been filled in but the FLW has not called NMS system once (via MA or MK).
- **Invalid FLW**: Records of an FLW are partly/fully present but are not valid any more due to CSV upload procedure marking the FLW as invalid.

In case an FLW whose record does not exist in NMS calls via IVR and subsequently, the record of user is updated via file upload (from any source – MCTS or training data), In such case, the status of FLW record shall reflect the changes done via file upload.

llavoie:
There are two small todos on #103 pending clarification on "mandatory" location fields and spec for "FLW Designation". I'll open another ticket for those in sprint 3 because we probably won't get clarification in time for sprint 2.


frankhuster:
lgtm 
assertTrue(a == b) => assertEquals(a, b)

PR #251, opened 2015-05-15
Fixes #122

### LINKED ISSUE #122: NMS.GEN.LANG.004 - language determination for user with unknown location
If the location information of FLW/MCTS-beneficiary is not known corresponding to the received MSISDN number, then following rules for language determination apply:
- **Case A**: If the circle information cannot be determined, the user is prompted in the national default language by the IVR to enter their Language-location code via DTMF.
- **Case B**: If the circle information can be determined and 
 - **Case B1**: there is only one “Language-location code” possible in the circle (based on geographical coverage of the circle), then the corresponding “Language-location code” shall be used for the received MSISDN number. 
 - **Case B2**: there are multiple languages possible in the circle (based on geographical coverage of the circle), the user is prompted in the circle’s default language by the IVR to enter their Language-location code via DTMF. 

Case B2 covers two types of cases. One case is of north-east where there are multiple states in a circle and consequently multiple languages. Another case is that where multiple languages are used within a state. Both cases are covered in this scenario.

larubbio:
There is still an open issue around Case B1 for which I'll put in a new ticket.

    Case B1: there is only one “Language-location code” possible in the circle (based on geographical  
    coverage of the circle), then the corresponding “Language-location code” shall be used for the 
    received MSISDN number.

The commit creates a national default language which is returned if no circle is provided.  It also returns all the valid languages for a circle if the user has no language preference saved.


koshalt:
lgtm

PR #260, opened 2015-05-18
Fixes #253

### LINKED ISSUE #253: WhitelistServiceImpl.whitelistEnabledForState
I think this is implemented wrong.  It looks like it is querying the whitelist_entry table and not the whitelisted_states table.  If that is the case then I don't think the _state table is ever used.

larubbio:
The check for whitelist being enabled for a state was looking in the wrong table.

### DIFF
--- a/flw/src/test/java/org/motechproject/nms/flw/osgi/WhiteListServiceBundleIT.java
+++ b/flw/src/test/java/org/motechproject/nms/flw/osgi/WhiteListServiceBundleIT.java
@@ -131,4 +131,38 @@ public class WhiteListServiceBundleIT extends BasePaxIT {
         boolean result = whitelistService.numberWhitelistedForState(whitelistState, NOT_WHITELIST_CONTACT_NUMBER);
         assertFalse(result);
     }
+
+    // Test state with whitelist, number not in list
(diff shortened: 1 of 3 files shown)


flw/src/test/java/org/motechproject/nms/flw/osgi/WhiteListServiceBundleIT.java
```diff
@@ -131,4 +131,38 @@ public void testStateWithWhitelistInvalidNumber() throws Exception {
         boolean result = whitelistService.numberWhitelistedForState(whitelistState, NOT_WHITELIST_CONTACT_NUMBER);
         assertFalse(result);
     }
+
+    // Test state with whitelist, number not in list
```
koshalt: does this comment need to be updated?

larubbio: yes. :(

llavoie: If it's easier, you can just update the comment using the github online editor and commit it.

PR #262, opened 2015-05-18
fixing small bug in Kilkari GetSubscriberDetails

llavoie:
We want GetSubscriberDetails to only return ACTIVE/PENDING subscriptions, not DEACTIVATED or COMPLETED.


larubbio:
lgtm

PR #264, opened 2015-05-19
CDR processing, third try

frankhuster:
Three phase CDR processing, some IT work remains to be done.


frankhuster:
**Not quite ready!**
I'm still working on decoupling the quick verifying of the file (handled during the POST from IMI) from the long verifying of the individual summary records (handled in a separate MOTECH event)

frankhuster:
I'll re-open when ready.

PR #266, opened 2015-05-20
CDR processing, take three

frankhuster:
Now a three phase process.


imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
```diff
@@ -17,7 +17,10 @@
...
     @Field
     @Column(allowsNull = "false")
-    private String status;
+    private Boolean success;
+
+    @Field
+    private String error;
 
     @Field
     private Integer recordCount;
```
larubbio: Are there any validations that can be added for this field?  Can it be null?  Must it be > 0?  Is there a max recordCount

imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
```diff
@@ -27,10 +30,12 @@
```
larubbio: Is there any validation for the checkSum field?  Should a length for the column be provided?

imi/src/main/java/org/motechproject/nms/imi/domain/FileProcessedStatus.java
```diff
@@ -4,23 +4,29 @@
...
 
     public int getValue() {
         return value;
     }
 
+    public String getName() {
+        return name;
+    }
+
     public static boolean isValid(int i) {
```
larubbio: When I read the method isValid I first thought it was returning if the File was valid based off of the status code.  Not a big deal, but I wonder if anyone else will confuse the enum itself being valid with the enum signifying the file is valid.

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
```diff
@@ -1,13 +1,44 @@
...
  */
 public interface CdrFileService {
 
-    ParseResults parseDetailFile(FileInfo fileInfo);
-    void processDetailFile(FileInfo fileInfo);
+    /**
+     * Internal method used to verify the given call detail record file (checksum, recordCount) and aggregate the
+     * detail records into a memory map of CallSummaryDetailDto objects collecting errors on the way
+     *
+     * NOTE: only exposed here for ITs
```
larubbio: I'm slightly concerned about methods exposed only for ITs.  It feels wrong.

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
```diff
@@ -1,13 +1,44 @@
...
+     */
+    VerifyResults verifyDetailFile(FileInfo fileInfo);
+
+
+    /**
+     * Aggregates multiple detail records provided my IMI into one summary record for each call in a given day.
+     * Then sends a MOTECH PROCESS_SUMMARY_RECORD event for each summary record such that the summary record
+     * process is distributed among all MOTECH nodes.
+     *
+     * NOTE: only exposed here for ITs. Normally called by the MOTECH event system (it's a @MotechListener)
```
larubbio: I'm slightly concerned about methods exposed only for ITs.  It feels wrong.

imi/src/main/java/org/motechproject/nms/imi/service/CsrValidatorService.java
```diff
@@ -0,0 +1,28 @@
...
+import org.motechproject.nms.kilkari.dto.CallSummaryRecordDto;
+
+import java.util.List;
+import java.util.Map;
+
+public interface CsrValidatorService {
+    /**
+     * Validates the provided call summary records, collecting errors on the go
+     *
+     * NOTE: directly used in IT only
```
larubbio: I'm slightly concerned about methods exposed only for ITs.  It feels wrong.

imi/src/main/java/org/motechproject/nms/imi/service/CsrValidatorService.java
```diff
@@ -0,0 +1,28 @@
...
+     * @param records a list of CallSummaryRecords
+     * @return a list of errors or an empty list
+     */
+    List<String> validateSummaryRecords(Map<String, CallSummaryRecordDto> records);
+
+
+    /**
+     * Validates a call summary record. Throws InvalidCallSummaryRecord exception if something goes wrong.
+
+     * NOTE: directly used in IT only
```
larubbio: I'm slightly concerned about methods exposed only for ITs.  It feels wrong.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,9 +57,14 @@
 @Service("cdrFileService")
 public class CdrFileServiceImpl implements CdrFileService {
 
+    private static final String CDR_FILE_NOTIFICATION_URL = "imi.cdr_file_notification_url";
     private static final String CDR_FILE_DIRECTORY = "imi.cdr_file_directory";
+    private static final String MAX_CDR_ERROR_COUNT = "imi.max_cdr_error_count";
+    private static final int MAX_CDR_ERROR_COUNT_DEFAULT = 100;
     private static final String PROCESS_SUMMARY_RECORD = "nms.imi.kk.process_summary_record";
```
larubbio: I was wondering why this key was different than the others (i.e. imi.* v. nms.imi.kk.*) I had to look through the code to see it was a subject, not a config key.  Can _SUBJECT or something be added to the constant to make that clear.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,9 +57,14 @@
...
     private static final String CSR_PARAM_KEY = "csr";
+    private static final String PROCESS_DETAIL_FILE = "nms.imi.kk.process_detail_file";
```
larubbio: Add _SUBJECT or something to make it clear that this is a subject, not a config key.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -52,36 +72,93 @@
...
         throw new IllegalStateException(error);
     }
 
 
-    // Aggregate all detail records for a requestId into a single summary record
-    private void parseDetailRecord(CallDetailRecordDto cdr, Map<String, CallSummaryRecordDto> records) {
+    private void reportAuditAndThrow(String file, List<String> errors) {
+        for (String error : errors) {
+            LOGGER.error(error);
+            alertService.create(file, "CDR Detail File", error, AlertType.CRITICAL, AlertStatus.NEW, 0, null);
```
larubbio: How will the end user clear these errors?  my concern is if a file is very broken that we'll place thousands of errors in the alert service.  Can the user bulk acknowledge?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -52,36 +72,93 @@
...
+        try {
+            HttpClient httpClient = HttpClients.createDefault();
+            HttpPost httpPost = new HttpPost(notificationUrl);
+            ObjectMapper mapper = new ObjectMapper();
+            String requestJson = mapper.writeValueAsString(cfpn);
+            httpPost.setHeader("Content-type", "application/json");
+            httpPost.setEntity(new StringEntity(requestJson));
+            HttpResponse response = httpClient.execute(httpPost);
+            int responseCode = response.getStatusLine().getStatusCode();
+            if (responseCode != HttpStatus.SC_OK) {
```
larubbio: We need to handle retry in case of timeout here and for all other API calls we make to IMI.  It is the last thing mentioned in the API doc.  We'll also need to add global config for the values (i.e. maxRetryAttempts etc.)

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -52,36 +72,93 @@
...
+        sendNotificationRequest(new CdrFileProcessedNotification(
+                FileProcessedStatus.FILE_ERROR_IN_FILE_FORMAT.getValue(),
+                file,
+                StringUtils.join(errors, ",")
+        ));
+    }
+
+
+    // Aggregate detail records for a requestId into single summary records using a (potentially huge) map
+    private void aggregateDetailRecord(CallDetailRecordDto cdr, Map<String, CallSummaryRecordDto> records) {
```
larubbio: I wonder if we will want to do some rough calculations on the memory usage.  A googleing says a java HashMap uses 32 * SIZE + 4 * CAPACITY bytes.  So each CallSummaryRecordDto has an internal map<Integer, Integer>, plus the large map of records.  Additionally the Dto itself will take up some amount of space.

We should certainly put a ticket in to test this in Sprint 4 at least.

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
frankhuster: Agreed, suggestions?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Yep, makes sense

frankhuster:
Good points thanks.
I'll try to address some and talk to you about how to deal with massive errors tomorrow.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -141,14 +229,21 @@ private ParseResults parseDetailRecords(File file, FileInfo fileInfo) {
 
             String line;
             while ((line = reader.readLine()) != null) {
-
                 try {
                     CallDetailRecordDto cdr = CsvHelper.csvLineToCdr(line);
 
-                    parseDetailRecord(cdr, records);
+                    if (!verifyOnly) {
```
larubbio: nit pick, but reading this as 'if not verifyOnly' confuses me.  I wonder if createing a local boolean at the type like this:

boolean provideAggregation = !verifyOnly

Then this if would read as 'if provideAggregation' which is less confusion (to me).  Either that or change the parameter but I'm not sure how that would read to the caller.  

Again this is a nit pick.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -121,18 +200,27 @@ private void parseDetailRecord(CallDetailRecordDto cdr, Map<String, CallSummaryR
...
+        try {
+            return Integer.parseInt(settingsFacade.getProperty(MAX_CDR_ERROR_COUNT));
+        } catch (NumberFormatException e) {
+            return MAX_CDR_ERROR_COUNT_DEFAULT;
+        }
+    }
+
+
+    @Override
+    public VerifyResults aggregateDetailFile(File file, FileInfo fileInfo, boolean verifyOnly) {
```
larubbio: Nit pick on naming.  This method is called 'aggregateDetailFile' but it returns a 'VerifyResults' object.  It also can be called such that it doesn't aggregate.

Maybe change the name to aggregateAndVerifyDetailFile?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -141,14 +229,21 @@ private ParseResults parseDetailRecords(File file, FileInfo fileInfo) {
...
+                        aggregateDetailRecord(cdr, records);
+                    }
 
                 } catch (IllegalArgumentException e) {
                     errors.add(String.format("Line %d: %s", lineNumber, e.getMessage()));
+                    errorCount++;
+                }
+                if (errorCount >= maxErrorCount) {
+                    errors.add(String.format("The maximum number of allowed errors of %d has been reached, " +
+                            "discarding all remaining errors.", maxErrorCount));
```
larubbio: Does it discard all remaining errors or stop processing the file?

frankhuster: It calls reportAuditAndThrow, which throws - so it ends processing here.

larubbio: Yes, so it doesn't "discard all remaining errors", it stops processing the file.  There may be many more errors, or there may be none.  We don't know since it stops processing.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -52,36 +72,93 @@
...
+                FileProcessedStatus.FILE_ERROR_IN_FILE_FORMAT.getValue(),
+                file,
+                StringUtils.join(errors, ",")
+        ));
+    }
+
+
+    // Aggregate detail records for a requestId into single summary records using a (potentially huge) map
+    private void aggregateDetailRecord(CallDetailRecordDto cdr, Map<String, CallSummaryRecordDto> records) {
         if (records.containsKey(cdr.getRequestId().toString())) {
```
larubbio: Completely premature optimization here.  I'm a little concerned that containsKey performance with a large map could be an issue.   How many records do we think we'll have in the map?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -90,6 +167,8 @@ private void parseDetailRecord(CallDetailRecordDto cdr, Map<String, CallSummaryR
                 record.setSecondsPlayed(cdr.getMsgPlayDuration());
             }
 
+            //todo: maybe verify the attemptCount provided by IMI and see if there are holes?
+
             /*
                Update the final status
```
larubbio: This comment is about line 178.

Can we accurately calculate the final status if the records aren't sorted?  If we get a success then we know that is the final status.  But all failure codes could be the actual final status and just because one comes later in a non-sorted file doesn't make it the final status.

There isn't anything we can do, just something we should note.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -157,39 +252,37 @@ private ParseResults parseDetailRecords(File file, FileInfo fileInfo) {
 
         } catch (IOException e) {
             String error = String.format("Unable to read %s: %s", file, e.getMessage());
-            reportErrorAndThrow(file.getName(), error);
+            reportAuditAndThrow(file.getName(), error);
         } catch (NoSuchAlgorithmException e) {
             String error = String.format("Unable to compute checksum: %s", e.getMessage());
-            reportErrorAndThrow(file.getName(), error);
+            reportAuditAndThrow(file.getName(), error);
```
larubbio: Should we throw here?  This doesn't seem like a bad request, it seems like an error on our end.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -197,12 +290,61 @@ private void sendProcessCdrEvent(CallSummaryRecordDto record) {
     }
 
 
+    // Phase 1: verify the file exists, the csv is valid and its record count and checksum match the provided info
+    //          while collecting a list of errors on the go.
+    //          Does not proceed to phase 2 if any error occurred and returns an error
+    @Override
+    public VerifyResults verifyDetailFile(FileInfo fileInfo) {
+        File userDirectory = new File(System.getProperty("user.home"));
+        File cdrDirectory = new File(userDirectory, settingsFacade.getProperty(CDR_FILE_DIRECTORY));
```
larubbio: I'm pretty sure this won't work once we get to integration testing with the IVR.  They can't drop the file into the home directory of the server that also happens to be processing the http post.

We need to get clarity around the location of the file and how we will access that server.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -197,12 +290,61 @@ private void sendProcessCdrEvent(CallSummaryRecordDto record) {
...
+    //          Validate all CallSummaryRecord fields exist in the database
+    //          Stop processing if errors occurred
     @Override
-    public void processDetailFile(FileInfo fileInfo) {
-        ParseResults parseResults = parseDetailFile(fileInfo);
+    @MotechListener(subjects = { PROCESS_DETAIL_FILE })
+    public List<String> processDetailFile(MotechEvent event) {
+        FileInfo fileInfo = (FileInfo) event.getParameters().get(FILE_INFO_PARAM_KEY);
+        File userDirectory = new File(System.getProperty("user.home"));
+        File cdrDirectory = new File(userDirectory, settingsFacade.getProperty(CDR_FILE_DIRECTORY));
```
larubbio: Same comment as above about the file location

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
larubbio: Thinking about this some more, and looking at the calling code instead of a boolean what about an enum of actions?

http://stackoverflow.com/questions/6282619/is-there-a-way-to-bitwise-or-enums-in-java

So instead of calling code like this:

```
aggregateDetailFile(file, fileInfo, false);
```

You can do this

```
aggregateDetailFile(file, fileInfo, EnumSet.of(Action.AGGREGATE, Action.VERIFY));
```

or

```
aggregateDetailFile(file, fileInfo, EnumSet.of(Action.VERIFY));
```

Then the signature would be:

```
public VerifyResults aggregateDetailFile(File file, FileInfo fileInfo, Set<Actions> actions)
```

and the check would be:

```
if (actions.contains(Actions.AGGREGATE) 
```

Not as concise as a bitflag but not terrible, and most likely not needed.  I just went down a rat hole, it's dark and scary in here.

imi/src/main/java/org/motechproject/nms/imi/web/ImiController.java
```diff
@@ -126,9 +132,13 @@ public void notifyNewCdrFile(@RequestBody CdrFileNotificationRequest request) {
...
 
-        cdrFileService.processDetailFile(request.getCdrDetail());
-    }
+        // Check the provided OBD file (aka: targetFile) exists in the FileAuditRecord table
+        verifyFileExistsInAuditRecord(request.getFileName());
 
+        // This checks the file, checksum, record count & csv lines, then sends an event to proceed to phase 2 of the
+        // CDR processing task also handled by the IMI module: processDetailFile
+        cdrFileService.verifyDetailFile(request.getCdrDetail());
+    }
```
larubbio: This method should be run in a transaction.

imi/src/main/java/org/motechproject/nms/imi/web/ImiController.java
```diff
@@ -153,17 +163,14 @@ public void notifyFileProcessedStatus(@RequestBody FileProcessedStatusRequest re
...
-        if (records.size() < 1) {
-            throw new NotFoundException("<fileName: Not Found>");
-        }
+        // Check the provided OBD file (aka: targetFile) exists in the FileAuditRecord table
+        verifyFileExistsInAuditRecord(request.getFileName());
 
-        // call OBD service, which will handle notification
+        //
         targetFileService.handleFileProcessedStatusNotification(request);
     }
```
larubbio: This method should be run in a transaction

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+            Map.Entry entry = (Map.Entry) it.next();
+            try {
+                validateSummaryRecord((CallSummaryRecordDto) entry.getValue());
+            } catch (InvalidCsrException e) {
+                errors.add(e.getMessage());
+
+            }
+            if (errorCount >= maxErrorCount) {
+                errors.add(String.format("The maximum number of allowed errors of %d has been reached, " +
+                        "discarding all remaining errors.", maxErrorCount));
```
larubbio: Instead of 'discarding all remaining errors' it is stopping the processing of the file.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+            }
+            weekIds.put(pack.getName(), packWeekIds);
+            contentFileNames.put(pack.getName(), packContentFileNames);
+        }
+    }
+
+
+    private void loadLanguageLocationCodes() {
+        languageLocationCodes = new HashSet<String>();
+        for (LanguageLocation languageLocation : languageLocationService.getAll()) {
```
larubbio: Since you just want the names, I wonder if we will need to create a method that returns just the language location codes.  My concern is since all of our related objects are in the defaultFetchGroup that to get all the codes we'll also have to load all the related objects (and maybe their related objects)  I have the same concerns about Circles.  This might be a Sprint 4 issue to test.

imi/src/main/java/org/motechproject/nms/imi/web/ImiController.java
```diff
@@ -103,6 +102,13 @@ private static boolean validateCdrFileInfo(StringBuilder errors, FileInfo fileIn
...
 
 
+    private void verifyFileExistsInAuditRecord(String fileName) {
+        List<FileAuditRecord> records =  fileAuditRecordDataService.findByFileName(fileName);
```
larubbio: If you are just looking for the existence of a record use the auto generate count() method.

imi/src/main/resources/imi.properties
```diff
@@ -17,6 +17,14 @@ imi.max_query_block=1000
...
 
+#url to use to notify the IVR system that the CDR file they sent was processed
+#4.3.2 CDRFileProcessedStatus Notification API
+imi.cdr_file_notification_url=http://localhost/obdmanager/NotifyCDRFileProcessedStatus
+
 #4.4.1 Target File Format
 #field #2 ServiceId - Unique Id provided by IMImobile for a particular service
-imi.target_file_imi_service_id=imiserviceid
\ No newline at end of file
+imi.target_file_imi_service_id=imiserviceid
```
larubbio: I thought the serviceId was going to change depending on the retry count for each record.  So fresh base calls would get serviceId 1.  First retry would get serviceId 2 and second retry some other service Id.  They would then do prioritization on their end based on the service id.

imi/src/main/resources/imi.properties
```diff
@@ -17,6 +17,14 @@ imi.max_query_block=1000
...
+
+#maximum number of errors allowed in a CDR file after which all errors are ignored so as to not overwhelm the
+#audit table & the tomcat log file
+imi.max_cdr_error_count=100
```
larubbio: We had discussed putting operations based config in files and other config in the db.  I'm not sure where this one falls.

imi/src/test/java/org/motechproject/nms/imi/it/CdrFileServiceBundleIT.java
```diff
@@ -89,18 +111,76 @@ public void testServicePresent() {
...
+        getLogger().debug("testVerify()");
+
+        CdrHelper helper = new CdrHelper(settingsService, subscriptionService, subscriberDataService,
+                languageDataService, languageLocationDataService, circleDataService, stateDataService,
+                districtDataService, fileAuditRecordDataService);
+
+        helper.makeCdrs(1,1,1,1);
+        helper.makeCdrFile();
+        FileInfo fileInfo = new FileInfo(helper.cdr(), helper.cdrChecksum(), helper.cdrCount());
+        Map<String, Object> eventParams = new HashMap<>();
```
larubbio: eventParams is never used

imi/src/test/java/org/motechproject/nms/imi/it/CdrFileServiceBundleIT.java
```diff
@@ -89,18 +111,76 @@ public void testServicePresent() {
...
+        CdrHelper helper = new CdrHelper(settingsService, subscriptionService, subscriberDataService,
+                languageDataService, languageLocationDataService, circleDataService, stateDataService,
+                districtDataService, fileAuditRecordDataService);
+
+        helper.makeCdrs(1, 1, 1, 1);
+        helper.makeCdrFile();
+        FileInfo fileInfo = new FileInfo(helper.cdr(), "invalid checksum", helper.cdrCount());
+        VerifyResults result = cdrFileService.verifyDetailFile(fileInfo);
+        assertEquals(0, result.getRecords().size()); //verifyDetailFile does not populate VerifyResults.records
+        assertEquals(1, result.getErrors().size());
```
larubbio: If the expectation of this test is an IllegalStateException then how come the last thing it does is an assert?  It seems like those lines wouldn't be reached.

imi/src/test/java/org/motechproject/nms/imi/it/CdrFileServiceBundleIT.java
```diff
@@ -89,18 +111,76 @@ public void testServicePresent() {
...
+        helper.makeCdrs(1,1,1,1);
+        helper.makeCdrFile();
+        FileInfo fileInfo = new FileInfo(helper.cdr(), helper.cdrChecksum(), helper.cdrCount());
+        Map<String, Object> eventParams = new HashMap<>();
+        VerifyResults result = cdrFileService.verifyDetailFile(fileInfo);
+        assertEquals(0, result.getErrors().size());
+    }
+
+
+    @Test(expected = IllegalStateException.class)
```
larubbio: Instead of using expected for the entire test I've been defining a rule so I can verify the exception is thrown on the line I expect it to be thrown from.

http://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests/2935935#2935935

imi/src/test/java/org/motechproject/nms/imi/it/CdrHelper.java
```diff
@@ -4,23 +4,19 @@
 import org.joda.time.DateTime;
```
larubbio: Arrrrrgggghhhhh TOO MUCH CODE TO REVIEW!!!!!  SOOOO BORED!!!!!

kilkari/src/test/java/org/motechproject/nms/kilkari/it/IntegrationTests.java
```diff
@@ -1,4 +1,4 @@
-package org.motechproject.nms.kilkari.osgi;
+package org.motechproject.nms.kilkari.it;
```
larubbio: Have we officially settled on 'it' over 'osgi'? :)

larubbio:
@fhuster It looks like this needs to have master merged in.  

Also when you delete subscribers/subscriptions in your tests, you'll need to add code like this to satisfy the pre-delete trigger:

```
        for (Subscription subscription: subscriptionDataService.retrieveAll()) {
            subscription.setStatus(SubscriptionStatus.COMPLETED);
            subscription.setEndDate(new DateTime().withDate(2011, 8, 1));

            subscriptionDataService.update(subscription);
        }
```

I've finished the review, look over the comments and see which ones you think are worthy or action. :see_no_evil:

imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
frankhuster: It's null when the entire file fails. But should probably be >= 0 when not null.

frankhuster: Any way to do that?

imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
frankhuster: Yep, 32

imi/src/main/java/org/motechproject/nms/imi/domain/FileProcessedStatus.java
frankhuster: now using `isValidEnumValue`

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
frankhuster: The alternative was to send the http request and wait 10 seconds :-(

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: yep

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: I setup a maximum number of errors after which the file will be rejected entirely and no more errors will be logged/audited/alerted

imi/src/main/java/org/motechproject/nms/imi/domain/FileAuditRecord.java
larubbio: What happens if you add @Min(0) to it?

koshalt:
you mean @frankhuster  right? @fhuster dunt eggcist

larubbio:
I think you responded on the wrong thread.

Anyway you can't possible expect me to remember all of the commoners names
can you?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Yep, implemented a configurable (non exponential) retry.

koshalt: @frankhuster any chance that this is in a util-like library? I need to do the same as well

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Sorting the incoming CDR and changing the algorithm should take care of that ;-)

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Since I'm now expecting a sorted CDR file I'm making things a bit more simple.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Yep, do you think that's a problem?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: See future checkin that'll say something like "cdr processing, take four: now sorting the incoming file"...

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
frankhuster: yep, now "ending file verification." instead.

imi/src/main/java/org/motechproject/nms/imi/web/ImiController.java
frankhuster: Ah, good point!

imi/src/test/java/org/motechproject/nms/imi/it/CdrHelper.java
frankhuster: I'm with you here...

frankhuster:
Of all my checkins this is the one I expected least to succeed... :confused:

frankhuster:
ok to test

frankhuster:
Just trying another time to see if I wasn't just lucky...

larubbio:
Funny, I've heard my parents say the same thing.

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
koshalt: Not to beat a dead horse, but these aren't tested via UTs (instead of ITs) because of all the class/config setup?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,167 +59,426 @@
...
+        try {
+            maxRetryCount = Integer.parseInt(settingsFacade.getProperty(MAX_NOTIFICATION_RETRY_COUNT));
+        } catch (NumberFormatException e) {
+            maxRetryCount = MAX_NOTIFICATION_RETRY_COUNT_DEFAULT;
+        }
+        int count = 0;
+
+        String error = "";
+
+        while (count < maxRetryCount) {
```
koshalt: is this supposed to have exponential backoff as well?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,167 +59,426 @@
...
             int statusCodeCount = 1;
-            if (record.getStatusStats().containsKey(cdr.getStatusCode())) {
-                statusCodeCount = record.getStatusStats().get(cdr.getStatusCode()) + 1;
+            if (csr.getStatusStats().containsKey(cdr.getStatusCode().getValue())) {
+                statusCodeCount = csr.getStatusStats().get(cdr.getStatusCode().getValue()) + 1;
             }
-            record.getStatusStats().put(cdr.getStatusCode().getValue(), statusCodeCount);
+            csr.getStatusStats().put(cdr.getStatusCode().getValue(), statusCodeCount);
 
             // Increment the message play duration
```
koshalt: this isn't a true increment right?  the multiple cdr seconds played don't get added?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Yep, good catch!

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
frankhuster: I chose IT because we're depending on the database & the file system.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: This is the number of times (not seconds) the call failed with a specific status.
Which point out the fact that this isn't properly documented, I'll fix that now.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,167 +59,426 @@
...
-            DigestInputStream dis = new DigestInputStream(fis, md);
+            DigestInputStream dis = null;
+            if (action == Action.PASS1) {
+                md = MessageDigest.getInstance("MD5");
+                dis = new DigestInputStream(fis, md);
+            }
 
             String line;
+            String currentRequestId = "";
+            CallSummaryRecordDto csr = null;
```
koshalt: I'm having trouble keeping up with all the different things that happen in this method. If I understand it correctly, the commonality between the passes is the iteration of the file, correct? 

I feel like the while loop should should be switch-cased on the pass or break it down into 3 different methods. I'm starting to side with the pmd on this one :cry:

frankhuster: I agree it's quite a complex method. I'll try to make it simpler.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,167 +59,426 @@
...
+    public List<String> processDetailFile(MotechEvent event) {
+        FileInfo fileInfo = (FileInfo) event.getParameters().get(FILE_INFO_PARAM_KEY);
+        File file = new File(localCdrDir(), fileInfo.getCdrFile());
+        String fileName = file.getName();
+
+        // Copy the file from the IMI network share (imi.remote_cdr_dir) into local cdr dir (imi.local_cdr_dir)
+        // Another MOTECH node than the one which picked up the HTTP cdrFileNotification might be picking up this
+        // event, so make sure to copy the file locally, if it's not there already.
+
+        if (file.exists()) {
```
koshalt: this condition would only be true when the remote and local cdr dir locations are the same, right? I don't remember seeing a copy operation on pass 1?

koshalt: nvm, i see that the controller calls the scp before pass 1. Would it make sense to have the scp helper as a separate method that gets invoked at different places?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -42,167 +59,426 @@
...
+        sendNotificationRequest(new CdrFileProcessedNotification(
+                FileProcessedStatus.FILE_PROCESSED_SUCCESSFULLY.getValue(),
+                fileInfo.getCdrFile(),
+                null
+        ));
+
+
+        // Phase 3: distribute the processing of individual CSRs to all nodes.
+        //          each node may generate an individual error that NMS-OPS will have to respond to.
+        errors = iterateDetailFile(file, fileInfo, Action.PASS3);
```
koshalt: if the max error count is set to n, does it only apply to each pass?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+        if (languageLocationCodes.contains(code)) {
+            return;
+        }
+        throw new InvalidCsrException(String.format("Invalid language location code: %s", code));
+    }
+
+
+    //todo: IT
+    @Override
+    public void validateSummaryRecord(CallSummaryRecordDto record) {
```
koshalt: no usages for this method yet? Just making sure...

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+                return;
+            }
+            throw new InvalidCsrException(String.format("Invalid contentFileName for %s pack: %s", pack, fileName));
+        }
+        throw new IllegalStateException(String.format("Unexpected pack name: %s", pack));
+    }
+
+
+    private void validateCircle(String name) {
+        if (circles == null) { loadCircles(); }
```
koshalt: this and other loads - how often does the cache get reloaded? Thinking about the  edge case where we get a new set of circles/llc etc from a data upload/sync followed by summary records.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -0,0 +1,181 @@
...
+            return Integer.parseInt(settingsFacade.getProperty(MAX_CDR_ERROR_COUNT));
+        } catch (NumberFormatException e) {
+            return MAX_CDR_ERROR_COUNT_DEFAULT;
+        }
+    }
+
+
+    //todo: IT
+    @Override
+    public List<String> validateSummaryRecords(Map<String, CallSummaryRecordDto> records) {
```
koshalt: no usages for this method yet? Just making sure...

imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
frankhuster: Good point, I imagine that any update of one of these basic tables should require a MOTECH restart.
But maybe we can have a 'refresh cache' event?

imi/src/main/java/org/motechproject/nms/imi/service/impl/SortHelper.java
```diff
@@ -0,0 +1,73 @@
...
+        return localFile;
+    }
+
+
+    /**
+     * Sorts the provided source CDR file on requestId (col1, text) and attempt (col4, int) into the destination
+     * file
+     */
+    void sort(String file) throws ExecException {
+        String command = String.format("%s -t , -k 1,1 -k 4,4n -o %s %s", getSortBinary(),
```
koshalt: nice!

imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
```diff
@@ -373,24 +354,19 @@ public TargetFileNotification generateTargetFile() {
...
-            recordCount = generateFreshCalls(today, maxQueryBlock, imiServiceId, callFlowUrl, targetFileName,
-                    writer);
+            recordCount = generateFreshCalls(today, maxQueryBlock, imiServiceId, callFlowUrl, writer);
 
             //Retry calls
-            recordCount += generateRetryCalls(today, maxQueryBlock, imiServiceId, callFlowUrl, targetFileName,
-                    writer);
+            recordCount += generateRetryCalls(today, maxQueryBlock, imiServiceId, callFlowUrl, writer);
 
             LOGGER.debug("Created targetFile with {} record{}", recordCount, recordCount == 1 ? "" : "s");
```
koshalt: grammar engineering :+1:

imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
```diff
@@ -422,13 +399,11 @@ private void sendNotificationRequest(TargetFileNotification tfn) {
                 String error = String.format("Expecting HTTP 200 response from %s but received HTTP %d : %s ",
```
koshalt: this needs retries too, correct?

imi/src/main/resources/imi.properties
```diff
@@ -4,11 +4,17 @@ imi.target_file_time=06:00
 #one day = 1000ms x 60s x 60m x 24h = 86400000
 imi.target_file_ms_interval=86400000
 
-#directory where target files should be generated (relative to the user's home)
-imi.target_file_directory=.motech/config/org.motechproject.nms.imi/target-files
+#local directory where target files should be generated
+imi.local_obd_dir=obb-files-local
```
koshalt: typo? obb-files-local

imi/src/test/java/org/motechproject/nms/imi/it/CdrHelper.java
koshalt: :+1:

imi/src/main/resources/imi.properties
frankhuster: Wow, good eye!!!

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/CallSummaryRecord.java
```diff
@@ -10,7 +10,7 @@
 import javax.validation.constraints.Min;
 import java.util.Map;
 
-@Entity(tableName = "nms_kilkari_call_summary_records")
+@Entity(tableName = "nms_kk_summary_records")
```
koshalt: unrelated but looks like we have a few different naming schemes we follow for the domain entity tables...even within the same module

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -253,10 +253,32 @@ public SubscriptionPack getSubscriptionPack(String name) {
...
+    @Override
+    public List<SubscriptionPack> getSubscriptionPacks() {
+        return subscriptionPackDataService.retrieveAll();
+    }
+
+
     /**
      * To be used by ITs only!
      */
     public void deleteAll() {
```
koshalt: :boom:

koshalt:
I didn't look too hard at the tests but the rest looks mostly good with a couple of items that need followup...either fixes or tracking issues, depending on what kind of a rush you are in.

frankhuster:
Thanks @koshalt I'll push the CR fixes soon & ping you.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
frankhuster: IKR!

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
koshalt: I might be missing something here. if there are multiple cdr(s) for the same csr, why would the play duration be added?

frankhuster: It's not a duration it's a count of the number of times the called failed,
for example we tried to call this number:

2 times there was a telco line error
3 times there was no pickup
1 time they picked up and immediately disconnected
1 time they picked up and we successfully played the message


On Tue, May 26, 2015 at 5:39 PM, kosh <notifications@github.com> wrote:

> In
> imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
> <https://github.com/motech-implementations/mim/pull/266#discussion_r31094172>
> :
>
> >              }
> > -            record.getStatusStats().put(cdr.getStatusCode().getValue(), statusCodeCount);
> > +            csr.getStatusStats().put(cdr.getStatusCode().getValue(), statusCodeCount);
> >
> >              // Increment the message play duration
>
> I might be missing something here. if there are multiple cdr(s) for the
> same csr, why would the play duration be added?
>
> —
> <https://github.com/motech-implementations/mim/pull/266/files#r31094172>.
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

koshalt: are we talking about the same thing? line 173? .getMsgPlayDuration?

frankhuster: We're most definitely not talking about the same thing.
I looked at line 173 and I now understand what you meant.
So I changed the comment to:
```java
/** Increment the message play duration
 *  The thinking here is that we're trying to capture the longest amount of message that was played
 *  for a call.
 */
if (cdr.getMsgPlayDuration() > csr.getSecondsPlayed()) {
    csr.setSecondsPlayed(cdr.getMsgPlayDuration());
}
```
Does that make more sense?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/CallSummaryRecord.java
frankhuster: yep, I imagine a project-wide table naming passthru wouldn't hurt...

PR #270, opened 2015-05-20
Fixes #268

### LINKED ISSUE #268: Handle subscription creation for subscriber with no language
Currently you can create a subscriber without a language.  I think MCTS upload is the path that allows for that.  If that happens I don't think there is a way to set their language.  What we need to do in SubscriptionServiceImpl#createSubscription is check if the subscriber's language is null, and if it is use the passed in language.

larubbio:
If there is a subscriber, but they do not have a language and the subscription does have a
language we assign that language to the user.

### DIFF
--- a/kilkari/src/test/java/org/motechproject/nms/kilkari/osgi/SubscriptionServiceBundleIT.java
+++ b/kilkari/src/test/java/org/motechproject/nms/kilkari/osgi/SubscriptionServiceBundleIT.java
@@ -300,6 +300,29 @@ public class SubscriptionServiceBundleIT extends BasePaxIT {
     }
 
     @Test
+    public void testCreateSubscriptionExistingSubscriberWithoutLanguage() throws Exception {
+        cleanupData();
+        createLanguageAndSubscriptionPacks();
+
+        LanguageLocation ta = languageLocationDataService.findByCode("10");
+
+        // Just verify the db is clean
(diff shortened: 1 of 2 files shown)


kilkari/src/test/java/org/motechproject/nms/kilkari/osgi/SubscriptionServiceBundleIT.java
```diff
@@ -300,6 +300,29 @@ public void testCreateSubscriptionExistingSubscriberDifferentLanguage() throws E
     }
 
     @Test
+    public void testCreateSubscriptionExistingSubscriberWithoutLanguage() throws Exception {
```
koshalt: is there a test for the negation of this one?...where the subscriber has a location  and so does the subscription, to make sure that it doesn't get overwritten?

larubbio: Yes, I think it is a few lines up in that file

koshalt: ok, I see testCreateSubscriptionExistingSubscriberDifferentLanguage

koshalt:
lgtm

PR #271, opened 2015-05-21
SCP module: calls unix scp and also sort

frankhuster:
Used to copy the CDR files from the IMI machine and sort it.
Also used to copy the OBD file to the IMI machine.

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
@@ -1,0 +1,86 @@
+package org.motechproject.nms.imi.service.impl;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+
+public final class SecureCopy {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SecureCopy.class);
+
+
+    private SecureCopy() { }
+
+
+    private static String getStderr(Process p) {
+        StringBuilder sb = new StringBuilder();
+        BufferedReader b = new BufferedReader(new InputStreamReader(p.getErrorStream()));
+        String line;
+
+        try {
+            while ((line = b.readLine()) != null) {
+                sb.append(line);
+                sb.append('\n');
+            }
+
+            b.close();
+        } catch (IOException e) {
+            LOGGER.error("Error trying to retrieve scp error message: {}", e.getMessage());
+        }
+        return sb.toString();
+    }
+
+
+    public static void copyFrom(String user, String host, String identityFile, String remoteSource,
+                                String localDestination) throws SecureCopyException {
+
+        Runtime r = Runtime.getRuntime();
+        try {
+            String identity;
+            if (identityFile == null || identityFile.isEmpty()) {
+                identity = "";
+            } else {
+                identity = String.format("-i %s ", identityFile);
+            }
+            String command = String.format("scp %s%s@%s:%s %s", identity, user, host, remoteSource, localDestination);
+            LOGGER.debug(command);
+            Process p = r.exec(command);
+            int retVal = p.waitFor();
+            if (retVal != 0) {
+                throw new SecureCopyException(String.format("Error copying from %s: scp returned %d - %s", host, retVal,
+                        getStderr(p)));
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new SecureCopyException(String.format("Error copying from %s: %s", host, e.getMessage()), e);
+        }
+    }
+
+
+    public static void copyTo(String user, String host, String identityFile, String localSource,
+                              String remoteDestination) throws SecureCopyException {
+
+        Runtime r = Runtime.getRuntime();
+        try {
+            String identity;
+            if (identityFile == null || identityFile.isEmpty()) {
+                identity = "";
+            } else {
+                identity = String.format("-i %s ", identityFile);
+            }
+            String command = String.format("scp %s%s %s@%s:%s", identity, localSource, user, host, remoteDestination);
+            LOGGER.debug(command);
+            Process p = r.exec(command);
+            int retVal = p.waitFor();
+            if (retVal != 0) {
+                throw new SecureCopyException(String.format("Error copying to %s: scp returned %d - %s", host, retVal,
+                        getStderr(p)));
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new SecureCopyException(String.format("Error copying to %s: %s", host, e.getMessage()), e);
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
```diff
@@ -0,0 +1,86 @@
...
+
+        Runtime r = Runtime.getRuntime();
+        try {
+            String identity;
+            if (identityFile == null || identityFile.isEmpty()) {
+                identity = "";
+            } else {
+                identity = String.format("-i %s ", identityFile);
+            }
+            String command = String.format("scp %s%s@%s:%s %s", identity, user, host, remoteSource, localDestination);
```
larubbio: I would use a fully qualified path to scp and perhaps make it a config value.  

Also is the command build correctly?  I thought to specify an identity file you would use scp -i %s %s@%s:%s %s  You need the -i and a space.

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
```diff
@@ -0,0 +1,86 @@
...
+            LOGGER.debug(command);
+            Process p = r.exec(command);
+            int retVal = p.waitFor();
```
larubbio: Does this wait indefinitely or does it timeout?

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
```diff
@@ -0,0 +1,86 @@
...
+
+        Runtime r = Runtime.getRuntime();
+        try {
+            String identity;
+            if (identityFile == null || identityFile.isEmpty()) {
+                identity = "";
+            } else {
+                identity = String.format("-i %s ", identityFile);
+            }
+            String command = String.format("scp %s%s %s@%s:%s", identity, localSource, user, host, remoteDestination);
```
larubbio: Same comment as above about full path and identity flag.

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
```diff
@@ -0,0 +1,86 @@
...
+            LOGGER.debug(command);
+            Process p = r.exec(command);
+            int retVal = p.waitFor();
```
larubbio: Same comment about timeout (which I think it does since you catch InterruptedException).  What is the default timeout?

imi/src/main/resources/imi.properties
```diff
@@ -19,4 +19,10 @@ imi.target_file_notification_url=http://localhost/obdmanager/notifytargetfile
...
 #4.4.1 Target File Format
 #field #2 ServiceId - Unique Id provided by IMImobile for a particular service
-imi.target_file_imi_service_id=imiserviceid
\ No newline at end of file
+imi.target_file_imi_service_id=imiserviceid
+
+
+# IMI file server scp information
+imi.scp.host=user
+imi.scp.user=p@ssw0rd
```
larubbio: Why is the username 'p@ssw0rd'.  :confused:

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+        Runtime r = Runtime.getRuntime();
+        try {
+            String identity;
+            if (identityFile == null || identityFile.isEmpty()) {
+                identity = "";
+            } else {
+                identity = String.format("-i %s ", identityFile);
+            }
+            String command = String.format("scp %s%s@%s:%s %s", identity, user, host, remoteSource, localDestination);
```
larubbio: Same comment as above.

Plus the additional comment of why is this code in both the imi and scp modules?

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new SecureCopyException(String.format("Error copying to %s: %s", host, e.getMessage()), e);
+        }
+    }
+
+
+    /**
+     * Sorts the source file into the destination file using the first 51 characters of the file as the sort key.
+     * The first 51 characters correspond to the requestId field in a CDR file.
+     */
+    void sort(String source, String destination) throws SortException {
```
larubbio: It feels wrong to have sort in a file called SecureCopy in a module called scp.

Also is sorting by the request id enough?  That will group all details for the same request, but it doesn't guarantee that they are in the order they were placed, which is what I think you want.

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        Runtime r = Runtime.getRuntime();
+        try {
+            String command = String.format("sort -o %s -k=1,52 %s", destination, source);
```
larubbio: I would use a full path to sort and externalize it as a config entry.

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
frankhuster: The space is on line 47.
Agree on the full path / config value (defaulting to /usr/bin/scp).

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
frankhuster: indefinitely

imi/src/main/resources/imi.properties
frankhuster: Because I wanted to see if you paid attention?

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
frankhuster: Yep, but I think it's easy to sort a handful of attemptIds for a given specific call.

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
frankhuster: Yep.

frankhuster:
Should I make this more of a platform module and call is something like the `os` module and then make the scp & sort more generic?
Or just move these methods to the imi module?

imi/src/main/java/org/motechproject/nms/imi/service/impl/SecureCopy.java
larubbio: ok.  I would just require the identityfile instead of making it optional.
 simpler code/

scp/src/main/java/org/motechproject/nms/scp/service/impl/SecureCopyServiceImpl.java
larubbio: true but I think sort can do it if you provide -k twice

larubbio:
if imi is the only consumer I would just move them there.

frankhuster:
OK, I'll kill the scp module and move things to the IMI module.
It'll be one happy large CR-feedback checkin.

frankhuster:
Moved scp & sort to the IMI module

PR #275, opened 2015-05-23
Fixes #132

### LINKED ISSUE #132: NMS.GEN.BKUP.002 - Data purging rules
The NMS system shall automatically remove all records marked Invalid (for FLW) or marked deactivated/completed (for beneficiary) from internal DB provided they satisfy the conditions listed in #131 NMS.GEN.BKUP.001.  

It should be possible to disable such automatic deletion.

larubbio:
Adds periodic purge of front line workers and beneficiaries


flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -103,8 +194,9 @@ public void deleteAllowed(FrontLineWorker frontLineWorker) {
             throw new IllegalStateException("FLW in invalid state with null invalidation date");
         }
 
-        if (Math.abs(Weeks.weeksBetween(now, frontLineWorker.getInvalidationDate()).getWeeks()) < 6) {
-            throw new IllegalStateException("FLW must be in invalid state for 6 weeks before deleting");
+        if (Math.abs(Weeks.weeksBetween(now, frontLineWorker.getInvalidationDate()).getWeeks()) < weeksToKeepInvalidFLWs) {
+            throw new IllegalStateException(String.format("FLW must be in %s state for %s weeks before deleting",
```
frankhuster: Why not use `"FLW must be in invalid state..."` instead of `"FLW must be in %s state...` just like in line 194?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -93,9 +182,11 @@ public void delete(FrontLineWorker record) {
 
     @Override
     public void deleteAllowed(FrontLineWorker frontLineWorker) {
```
frankhuster: The name is misleading, I would assume the this would return `true` if it's possible to delete the FLW.

flw/src/main/resources/flw.properties
```diff
@@ -0,0 +1,7 @@
+#what time of day should the process of generating the targetFile start
```
frankhuster: Why?

flw/src/main/resources/flw.properties
```diff
@@ -0,0 +1,7 @@
...
+flw.purge_invalid_flw_start_time=06:00
+
+#one day = 1000ms x 60s x 60m x 24h = 86400000
+flw.purge_invalid_flw_ms_interval=86400000
```
frankhuster: I'm uncomfortable specifying a day interval in seconds (I do the same in the IMI module)
It feels like we should be able to explicitly schedule an event daily at a specific hour.
Opinion?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
larubbio: I originally did that but then decided I wanted to log exactly the same as what I was comparing against.  I should change line 194

flw/src/main/resources/flw.properties
larubbio: I just copied your code from the cdr scheduling.  Do you mean I should change the comment or remove the ability to set the time?

frankhuster: I was wondering why the discarding of old FLW should coincide with targetFile generation

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
larubbio: ok, changed to deletePreconditionCheck

flw/src/main/resources/flw.properties
larubbio: It doesn't, I just didn't update the comment that I pulled from your config file.

flw/src/main/resources/flw.properties
larubbio: I was thinking of using the batch module for this, but ended up thinking it was overkill.  It might make sense to look into it again.

I do think having sub-day granularity is useful (at least for the data purge).  If there is a lot of data to purge running it twice a day could be useful.

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -17,11 +30,87 @@
...
+
+                query.setFilter("status == invalid && invalidationDate < cutoff");
+                query.declareParameters("org.motechproject.nms.flw.domain.FrontLineWorkerStatus invalid, org.joda.time.DateTime cutoff");
+
+                return query.deletePersistentAll(status, cutoff);
+            }
+        };
+
+        Long purgedRecordCount = frontLineWorkerDataService.executeQuery(queryExecution);
+        LOGGER.info(String.format("Purged %s FLWs with statis %s and invalidation date before %s FLWs",
```
frankhuster: Change to:
`"Purged %s FLWs with status %s and invalidation date before %s"`
notice stat**u**s and missing trailing FLWs

larubbio: Arrête de te la péter

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -17,11 +30,87 @@
...
+
+        @SuppressWarnings("unchecked")
+        QueryExecution<Long> queryExecution = new QueryExecution<Long>() {
+            @Override
+            public Long execute(Query query, InstanceSecurityRestriction restriction) {
+
+                query.setFilter("status == invalid && invalidationDate < cutoff");
+                query.declareParameters("org.motechproject.nms.flw.domain.FrontLineWorkerStatus invalid, org.joda.time.DateTime cutoff");
+
+                return query.deletePersistentAll(status, cutoff);
```
frankhuster: Now is probably not the time (Sprint 4 might be) but wouldn't you like to have an index on invalidationDate & status?

larubbio: We should have one on status for sure.  If there is one on status and
invalidationDate then it would work for both cases.  I'll add it.

frankhuster:
lgtm

Issue #218, opened 2015-05-06
Upload languageLocationCodes


larubbio:
Issue #119 has several features within it.  This one should be handled as it's own task:

> There shall be one and only one “Language-location code” used in a given geographical location.
> This “Language-location code” shall be same and applicable to all services in that location.

llavoie:
@larubbio please add CSV schema info

larubbio:
![screen shot 2015-05-06 at 11 11 01 am](https://cloud.githubusercontent.com/assets/51527/7618354/86a9ea12-f966-11e4-93cd-4a50c84a25b4.png)

larubbio:
Below is data from the sample file that was mailed to me over the weekend.  I think we need to modify the domain entity so there is a many to many relationship between circle and language location code.  I'm also confused about the value of this table.

Circle | State | District | languagelocation code | Language | Default Language for Circle (Y/N)
-------|-------|----------|-----------------------|----------|----------------------------------
Delhi - NCR | Delhi | North District | HIN | Hindi | Y
Delhi - NCR | Delhi | Central District | HIN | Hindi | Y
Delhi - NCR | Delhi | South District | HIN | Hindi | Y
Delhi - NCR | Delhi | New Delhi District | HIN | Hindi | Y
Delhi - NCR | Delhi | West District | HIN | Hindi | Y
Delhi - NCR | Delhi | North East District | HIN | Hindi | Y
Delhi - NCR | Delhi | East District | HIN | Hindi | Y
Delhi - NCR | Delhi | North West District | HIN | Hindi | Y
Delhi - NCR | Delhi | South East District | HIN | Hindi | Y
Delhi - NCR | Delhi | SHAHDARA | HIN | Hindi | Y
Delhi - NCR | Delhi | Shahdara District | HIN | Hindi | Y
UPEast | Uttar Pradesh | Allahabad | HIN | Hindi | Y
UPEast | Uttar Pradesh | Azamgarh | HIN | Hindi | Y
UPEast | Uttar Pradesh | Bahraich | HIN | Hindi | Y
UPEast | Uttar Pradesh | Ballia | HIN | Hindi | Y
UPEast | Uttar Pradesh | Balrampur | HIN | Hindi | Y
UPEast | Uttar Pradesh | Banda | HIN | Hindi | Y
UPEast | Uttar Pradesh | Barabanki | HIN | Hindi | Y
UPEast | Uttar Pradesh | Basti | HIN | Hindi | Y
[...]

PR #279, opened 2015-05-26
#218: Language Location Codes upload

kamknitt:


llavoie:
lgtm, but I'd also like @larubbio to weigh in because he knows much more than I do about the language location code functional requirements.

BTW, I'm probably going to want to extract the region/utils directory into its own module so that we can reuse those utility classes in other places (e.g. MCTS import -- and I assume you'll want those for FLW import).

larubbio:
I updated the ticket with the sample data I received.  I think we'll have to change the datamodel to accommodate it which means we'll need some changes to this as well.

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+public class LanguageLocationCodesImportServiceImpl implements LanguageLocationCodesImportService {
+
+    public static final Logger LOGGER = LoggerFactory.getLogger(LanguageLocationCodesImportServiceImpl.class);
+
+    public static final String LANGUAGE_LOCATION_CODE = "languageLocationCode";
+    public static final String LANGUAGE = "language";
+    public static final String CIRCLE = "circle";
+    public static final String STATE = "state";
+    public static final String DISTRICT = "district";
+    public static final String DEFAULT_FOR_CIRCLE = "defaultForCircle";
```
larubbio: From the sample file I think the column headers will be:

Circle | State | District | languagelocation code | Language | Default Language for Circle (Y/N)
--------|---------|-----------|-------------------------------|---------------|----------------------------------------------

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private void importRecord(Map<String, Object> record) {
+        String languageLocationCode = (String) record.get(LANGUAGE_LOCATION_CODE);
+        Language language = (Language) record.get(LANGUAGE);
+        State state = (State) record.get(STATE);
+        District district = (District) record.get(DISTRICT);
+        Circle circle = (Circle) record.get(CIRCLE);
+        Boolean defaultForCircle = (Boolean) record.get(DEFAULT_FOR_CIRCLE);
+
+        if (null != state && null != district) {
```
larubbio: I think the import processor rejects rows with null state or district.

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+    private void importRecordForDistrict(String languageLocationCode, Language language, Circle circle,
+                                         Boolean defaultForCircle, District district) {
+        LOGGER.warn("State not specified for the '{}' district", district.getName());
+        verifyStateInCircle(circle, district.getState());
+        verifyDistrictLanguageLocation(district);
+        addDistrictToLanguageLocation(languageLocationCode, language, circle, defaultForCircle, Arrays.asList(district));
+    }
+
+    private void addDistrictToLanguageLocation(String languageLocationCode, Language language, Circle circle, Boolean defaultForCircle, Collection<District> districts) {
+        LanguageLocation languageLocation = languageLocationDataService.findByCode(languageLocationCode);
```
larubbio: Looking at the sample file languageLocationCode isn't actually unique.  I think if we change the domain object so circle <-> languageLocation becomes many to many than it can be unique again.   We'll have to make that change soon.

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+        if (null != languageLocation) {
+            verify(Objects.equals(languageLocation.getLanguage().getName(), language.getName()),
+                    "Language location exists, yet its language does not match the language supplied in the CSV record");
+            verify(Objects.equals(languageLocation.getCircle().getName(), circle.getName()),
+                    "Language location exists, yet its circle does not match the circle supplied in the CSV record");
```
larubbio: This will have to change to a contains check or be removed once the relation is switched to many to many

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+            public State retrieve(Long value) {
+                State state = stateDataService.findByCode(value);
+                verify(null != state, "State does not exists");
+                return state;
+            }
+        }));
+        mapping.put(DISTRICT, new Optional(new GetInstanceByLong<District>() {
+            @Override
+            public District retrieve(Long value) {
+                District district = districtDataService.findByCode(value);
```
larubbio: The sample file they gave me uses district name, not the code. :(

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+            public Circle retrieve(String value) {
+                Circle circle = circleDataService.findByName(value);
+                verify(null != circle, "Circle does not exists");
+                return circle;
+            }
+        });
+        mapping.put(STATE, new Optional(new GetInstanceByLong<State>() {
+            @Override
+            public State retrieve(Long value) {
+                State state = stateDataService.findByCode(value);
```
larubbio: The sample file they gave me has a hidden column that is blank for state code.  It does have the state name.

region/src/main/java/org/motechproject/nms/region/utils/GetBoolean.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class GetBoolean implements CellProcessor {
+
+    @Override
+    public Object execute(Object value, CsvContext context) {
+        Boolean returnValue;
+        if (value instanceof Boolean) {
+            returnValue = (Boolean) value;
+        } else if (value instanceof String) {
+            returnValue = Boolean.valueOf((String) value);
```
larubbio: The sample file they sent uses 'Y' and 'N' for boolean values.  (I added the file to the ticket)

region/src/test/java/org/motechproject/nms/region/ut/CsvImportTest.java
```diff
@@ -35,90 +36,92 @@
...
     public void testReadSampleWhenProcessorConstraintIsViolated() throws Exception {
+        CsvInstanceImporter<Sample> csvInstanceImporter = createCsvInstanceImporter();
         Reader reader = new StringReader("n,s,o\n12,,1");
-        csvImporter.open(reader, getFieldNameMapping(), getProcessorMapping());
-        csvImporter.read();
+        csvInstanceImporter.open(reader, getProcessorMapping(), getFieldNameMapping());
+        csvInstanceImporter.read();
     }
 
     @Test
```
larubbio: Should a test for GetBoolean be added since we'll probably want to allow true/false and y/n as booleans.

larubbio:
@kamknitt I think the column names need to be updated and the boolean field changed to allow y/n.  and then it looks ok.  Although now that I've seen the sample data file I think we'll have to update the language location entity and then make changes to this.  So maybe merge but leave the ticket open?

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
kamknitt: Please note that state and district processors are wrapped in the Optional processor. It causes the column with the empty value to resolve to null without calling the wrapped processor. For example, the exception will be thrown from processor only when the state code is provided and the state with this code does not exist.

kamknitt:
@larubbio I've fixed issues with column names and boolean. Merging what we have now and leaving ticket open sounds good to me.
Do you want me to take care of updating language location entity as a part of this ticket?

region/src/main/java/org/motechproject/nms/region/service/impl/LanguageLocationCodesImportServiceImpl.java
larubbio: Sounds good.  I was wondering how the tests were passing

larubbio:
ok.  I'll merge it.  For the changes to language location, I was thinking of just getting rid of it.

We need to track the relationship between circles and states.  (which I think we already do)  For each circle we should add the default language.

Additionally in District we should add language.

If we do that and then update the api calls to walk the relation ship then I think everything should work.  

In FrontLineWorker we can should change LanguageLocation to just be language.  (There is a story about how changing the users location shouldn't change their language)

In Subscriber we should also change LanguageLocation to Language.

PR #283, opened 2015-05-27
Course content

koshalt:
Adding course resource json and moving away from domain objects


mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/MobileAcademyService.java
```diff
@@ -1,7 +1,7 @@
 package org.motechproject.nms.mobileacademy.service;
 
-import org.motechproject.nms.mobileacademy.domain.Course;
 import org.motechproject.nms.mobileacademy.dto.MaBookmark;
+import org.motechproject.nms.mobileacademy.dto.MaCourse;
 
 /**
  * Simple example of a service interface.
```
llavoie: Hello World turd to clean up...

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -28,6 +36,10 @@
 @Service("mobileAcademyService")
 public class MobileAcademyServiceImpl implements MobileAcademyService {
 
+    private static final String COURSE_CONTENT_FILE = "nmsCourse.json";
+
+    private static final String COURSE_NAME = "MobileAcademyCourse";
+
     private static final String FINAL_BOOKMARK = "Chapter11_Quiz";
```
llavoie: Is there no spec requirement for these values to be configurable? It would be unfortunate if the course was changed to have 12 chapters and that required a recompile due to hard-coded constants...

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -51,99 +63,111 @@
...
+    public void setCourse(MaCourse courseDto) {
 
-        if (existing == null) {
-            courseDataService.create(course);
-        } else {
-            course.setId(existing.getId());
-            courseDataService.update(course);
+        if (courseDto == null) {
+            LOGGER.error("Attempted to set null course, exiting operation");
+            alertService.create("MA.Course", "MaCourse", "Trying to set null MaCourse", AlertType.CRITICAL, AlertStatus.NEW, 0, null);
```
llavoie: It looks like some of the alert subjects are MA.Course and some are mTraining.Course. Intentional?

koshalt: Nice catch. I missed this during the refactor away from mtraining course.
Changed to MA.Course now

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
koshalt: There is a spec requirement ..but the other way. There is guaranteed to
only ever be 11 chapters and I believe the vxml parsing relies on this
'magic' number for their looping logic.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/MobileAcademyService.java
koshalt: turd bin clindup :hankey:

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -51,99 +63,111 @@
...
-            if (existing.size() > 1) {
-                LOGGER.error("Found more than 1 bookmark for calling number. This should never be possible.");
-                LOGGER.error("Contact dev team about calling number: " + callingNumber);
-            }
 
             // update the first bookmark
-            LOGGER.info("Updating the first bookmark for user");
-            bookmarkDataService.update(setBookmarkProperties(saveBookmark, existing.get(0)));
+            LOGGER.info("Updating bookmark for user");
+            bookmarkService.updateBookmark(setBookmarkProperties(saveBookmark, existingBookmark));
```
llavoie: On this line we set the bookmark properties to those of the saveBookmark, but then just below we check whether it's null. Should we be checking that earlier?

koshalt: you are right. That would be a bad request exception that needs to be checked in the api controller as well. I'll update and push out another patch for this.

mobile-academy/src/test/java/org/motechproject/nms/mobileacademy/it/MobileAcademyServiceBundleIT.java
```diff
@@ -61,16 +63,49 @@
...
+
+    @Test
+    public void testSetCourseUpdate() {
+
+        NmsCourse originalCourse = nmsCourseDataService.getCourseByName("MobileAcademyCourse");
+        String courseContent = originalCourse.getContent();
+        MaCourse copyCourse = new MaCourse(originalCourse.getName(), originalCourse.getModificationDate().getMillis(), originalCourse.getContent() + "foo");
+        maService.setCourse(copyCourse);
+
+        // verify that modified time (version) didn't change
```
llavoie: Nitpick: the comment should say that the time DID change, right?

koshalt: oops..copy-paste error :grin: . will fix

llavoie:
OK, that's all from me -- looks good aside from my small comments.

PR #285, opened 2015-05-29
IT refactor

frankhuster:
All Integration Tests are now run from the testing module.

This first PR has the tests more or less me the same except they're all running from one module inside one suite. The next step is to see if there is some refactoring to do and similar code to delete.


frankhuster:
ok to test

frankhuster:
ok to test

flw/src/main/resources/META-INF/spring/blueprint.xml
```diff
@@ -33,6 +33,9 @@
     <osgi:service ref="whitelistService"
                   interface="org.motechproject.nms.flw.service.WhitelistService" />
 
+    <osgi:service ref="itService"
```
koshalt: nothing to do now but should we consider adding an ops task to 'unwire' the itService dependency for production?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/InboxCallDetailRecord.java
```diff
@@ -47,11 +48,14 @@
...
 
-    public InboxCallDetailRecord() { }
+    public InboxCallDetailRecord() {
+        content = new HashSet<>();
+    }
 
     public InboxCallDetailRecord(Long callingNumber, String operator, String circle, // NO CHECKSTYLE More than 7 parameters
                                  Long callId, DateTime callStartTime, DateTime callEndTime, Integer callDurationInPulses,
                                  Integer callStatus, Integer callDisconnectReason, Set<InboxCallData> content) {
+        this();
```
koshalt: why do this instead of just initializing content here?

flw/src/main/resources/META-INF/spring/blueprint.xml
frankhuster: Or something like having the it service disable itself from a settings?
So we don't have to recompile

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/InboxCallDetailRecord.java
frankhuster: good practice to always use the base ctor instead of duplicating code.

koshalt: I'm with you on the base constructor for inheritance and the this(...) in the sense of calling a parameterized constructor from a no param constructor - like this(a, 10) from no param.

but this seems a little backwards to me in the sense of calling a no param constructor from a parameterized one and splitting up the initialization between 2 constructors.

Not a deal breaker though.

frankhuster: Good point, I'll do it the other way when I refactor the data seeding code.

props/src/main/java/org/motechproject/nms/props/domain/FinalCallStatus.java
```diff
@@ -19,13 +22,13 @@ public int getValue() {
...
     public static boolean isValidEnumValue(int i) {
-        return (i == 1 || i ==2 || i == 3);
+        return (i == SU || i == FA || i == RE);
     }
 
     public static FinalCallStatus fromInt(int i) {
-        if (i == 1) { return SUCCESS; }
-        if (i == 2) { return FAILED; }
-        if (i == 3) { return REJECTED; }
+        if (i == SU) { return SUCCESS; }
```
koshalt: If you read this fast enough, it says sufare..like safari

props/src/main/java/org/motechproject/nms/props/domain/FinalCallStatus.java
```diff
@@ -19,13 +22,13 @@ public int getValue() {
...
+        if (i == FA) { return FAILED; }
+        if (i == RE) { return REJECTED; }
```
koshalt: also, this seems like a standard performant way to get enum from int. http://stackoverflow.com/questions/11047756/getting-enum-associated-with-int-value

No need to change anything, just googling around on a friday and trying to write performance enhancing code.

Oh and switch-case :see_no_evil: is so much better than if, according to studies by me.

In other news, why did I go to college when I could have learnt everything from stack overflow in 4yrs?

testing/pom.xml
```diff
@@ -0,0 +1,329 @@
...
+                                <goals>
+                                    <goal>integration-test</goal>
+                                </goals>
+                                <configuration>
+                                    <includes>
+                                        <include>**/**/*IT.java</include>
+                                        <include>**/**/*IntegrationTests.java</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>**/**/*BundleIT.java</exclude>
```
koshalt: do you know why we only stick to these naming conventions to pick up the tests?

testing/pom.xml
```diff
@@ -0,0 +1,329 @@
...
+                            <execution>
+                                <id>create-db</id>
+                                <phase>pre-integration-test</phase>
+                                <goals>
+                                    <goal>execute</goal>
+                                </goals>
+                                <configuration>
+                                    <autocommit>true</autocommit>
+                                    <sqlCommand>
+                                        CREATE DATABASE motech_data_services;
```
koshalt: we should also specify the default character set here to be explicity since that varies by the mysql version

testing/pom.xml
```diff
@@ -0,0 +1,329 @@
...
+                            org.motechproject.nms.flw.service,
+                            org.motechproject.nms.flw.repository,
+                            org.motechproject.nms.flw.domain,
+                            org.motechproject.nms.props.service,
+                            org.motechproject.nms.imi.service,
+                            org.motechproject.nms.region.service,
+                            org.motechproject.nms.region.repository,
+                            org.motechproject.nms.region.domain,
+                            org.motechproject.alerts.contract,
+                            *
```
koshalt: no ma imports?

testing/src/test/java/org/motechproject/nms/testing/it/api/MobileAcademyControllerBundleIT.java
```diff
@@ -147,7 +141,7 @@ public void testGetCourseValid() throws IOException, InterruptedException {
         HttpGet request = RequestBuilder.createGetRequest(endpoint);
 
         HttpResponse httpResponse = SimpleHttpClient.httpRequestAndResponse(request, RequestBuilder.ADMIN_USERNAME, RequestBuilder.ADMIN_PASSWORD);
-        assertTrue(httpResponse.getStatusLine().getStatusCode() == 200);
+        assertEquals(200, httpResponse.getStatusLine().getStatusCode());
```
koshalt: not a big fan of this style personally. I like to assert on the boolean of my expected assumption. oh well.

testing/src/test/java/org/motechproject/nms/testing/it/api/utils/SubscriptionPackBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final int TWO_MINUTES = 120;
+    private static final int TEN_SECS = 10;
+
+    public static SubscriptionPack createSubscriptionPack(String name, SubscriptionPackType type, int weeks,
+                                                          int messagesPerWeek) {
+        List<SubscriptionPackMessage> messages = new ArrayList<>();
+        for (int week = 1; week <= weeks; week++) {
+            messages.add(new SubscriptionPackMessage(week, String.format("w%s_1", week),
+                    String.format("w%s_1.wav", week),
+                    TWO_MINUTES - TEN_SECS + (int) (Math.random() * 2 * TEN_SECS)));
```
koshalt: not entirely sure what's going on here. what is this math for?

koshalt:
minor comments. lgtm.

testing/src/test/java/org/motechproject/nms/testing/it/api/utils/SubscriptionPackBuilder.java
frankhuster: I want roughly 2 minutes long messages but not all excatly the same:  +- 10s

testing/src/test/java/org/motechproject/nms/testing/it/api/MobileAcademyControllerBundleIT.java
frankhuster: But the problem here is you don't know what the returned value was, you just know it wasn't 200.
This tells you something like "expected 200 but received 404"
~is better no?~

testing/pom.xml
frankhuster: yep

testing/pom.xml
frankhuster: Nope, Pawel?

flw/src/main/resources/META-INF/spring/blueprint.xml
koshalt: I thought blueprint was runtime wiring config anyways. Maybe not

testing/src/test/java/org/motechproject/nms/testing/it/api/MobileAcademyControllerBundleIT.java
koshalt: :confounded: yes yes. is better. although, my tests never fail :grinning:

frankhuster:
ok to test

frankhuster:
ok to test

PR #291, opened 2015-06-01
Refactor generic CSV code from Region module into its own module

llavoie:
Pulling this code out so that it can be used by Kilkari and FLW for CSV import work items.


region/pom.xml
```diff
@@ -85,9 +91,10 @@
                             org.motechproject.nms.region.service;version=${project.version},
                             org.motechproject.nms.region.repository;version=${project.version},
                             org.motechproject.nms.region.domain;version=${project.version},
-                            org.motechproject.nms.region.exception;version=${project.version}
                         </Export-Package>
                         <Import-Package>
+                            org.motechproject.nms.csv.exception,
```
larubbio: Are these not explicitly referenced in the code?  I'm surprised they need to be added here.

llavoie: You're right, this isn't needed.

llavoie:
ok to test

PR #293, opened 2015-06-01
FT140

frankhuster:
Initial commit with the new IT infrastructure


frankhuster:
In addition to starting the IT move to the new infrastructure, @srivastavaAbhi and I also fixed a couple of build settings (checkstyle & removed some unnecessary files)

frankhuster:
ok to test

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/CsrServiceBundleIT.java
```diff
@@ -339,6 +339,58 @@ public void verifySubscriptionCompletion() {
         assertEquals(3, subscriptions.size());
     }
 
+
+    @Test
+    public void verifyFT140() {
```
koshalt: I think if we are sticking with the test#id for the naming, that we have a short description for what the test is trying to do (instead of cross-referencing the code and spec all the time)

llavoie: Agreed -- either the test case should have a descriptive name and a comment with the id #, or the test can be named with the id # but should have a descriptive comment. 

-----Original Message-----

frankhuster: Yep, that's eaxctly what Abhishek and I are doing right now.

testing/src/test/java/org/motechproject/nms/testing/it/imi/TargetFileServiceBundleIT.java
```diff
@@ -199,13 +188,42 @@ public void restoreSettings() {
...
+        Subscriber subscriber2 = new Subscriber(2222222222L, urdu, bb);
+        subscriber2.setLastMenstrualPeriod(DateTime.now().minusDays(90)); // startDate will be today
+        subscriberDataService.create(subscriber2);
+        Subscription subscription2 = subscriptionService.createSubscription(2222222222L, urdu, pregnancyPack,
+                SubscriptionOrigin.MCTS_IMPORT);
+        subscriptionService.deactivateSubscription(subscription2, DeactivationReason.CHILD_DEATH);
+
+        //Should not be picked up because it's not for today
+        Subscriber subscriber3 = new Subscriber(6666666666L, urdu, bb);
+        subscriber3.setDateOfBirth(DateTime.now()); // startDate is DOB + 1 for child packs, so setting the DOB
```
koshalt: maybe an edge case - but this test could fall on the day boundary and cause failures as we roll on to dob+1 ? Can we set this to DateTime.now() + 1?

koshalt:
lgtm. waiting for update to merge

koshalt:
test this please

Issue #101, opened 2015-04-25
NMS.GEN.FLW.001 - FLW Data Import


larubbio:
The NMS system shall provide means to upload FLW data into its internal database via a web interface via a CSV file (or web service).

Multiple upload attempts shall be supported whereby the complete FLW data may not be available in the beginning.

Errors during FLW data upload shall be logged.

koshalt:
Is this still the plan of action with the data service from MOH coming up? If yes, I think we still have a gap in how the csv is authored or brought to the ops people...a can that we've been kicking down the road for a few months now

llavoie:
Example of FLW data: 

https://gist.github.com/llavoie/fbe2106baed00746ec8e

llavoie:
Fixed by @kamknitt

PR #295, opened 2015-06-03
#101: FLW Data Import

kamknitt:


flw/src/main/java/org/motechproject/nms/flw/controller/FrontLineWorkerImportController.java
```diff
@@ -0,0 +1,62 @@
...
+@Controller
+public class FrontLineWorkerImportController {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FrontLineWorkerImportController.class);
+
+    private AlertService alertService;
+
+    private FrontLineWorkerImportService frontLineWorkerImportService;
+
+    @RequestMapping(value = "/flw/import", method = RequestMethod.POST)
```
larubbio: Will flw be duplicated in the url since the module name is flw?

testing/src/test/java/org/motechproject/nms/testing/it/flw/FrontLineWorkerImportServiceBundleIT.java
```diff
@@ -0,0 +1,165 @@
...
+        healthSubFacilityDataService.deleteAll();
+        healthFacilityDataService.deleteAll();
+        healthFacilityTypeDataService.deleteAll();
+        healthBlockDataService.deleteAll();
+        villageDataService.deleteAll();
+        talukaDataService.deleteAll();
+        districtDataService.deleteAll();
+        stateDataService.deleteAll();
+        circleDataService.deleteAll();
+
```
larubbio: Instead of calling all of the deleteAll you can call TestingServiceImpl.clearDatabase()

larubbio:
Small comments, in general looks good to me.

I do think we'll need a new ticket since an FLWs location has to be provided at the District level, but it may be provided at a more granular level.  If it is then we are supposed to save it at that lower level of detail.

llavoie:
A few more questions:

- The sample data we have from the government web service (https://gist.github.com/llavoie/fbe2106baed00746ec8e) is tab-delimited, so don't we need to use CsvPreference.TAB_PREFERENCE here?
- The sample data also has a few junk lines at the top of the file -- will those be skipped?
- The sample data also has quite a few more columns than the ones that this code is processing -- will the CsvImporter skip those or do they need to be handled somewhere? e.g. do we need to include them in the ProcessorMapping even if they don't get saved on the FLW record?

frankhuster:
I think we should have an IT that uses the same file (not just a
StringReader object) Lauren provided:
https://gist.github.com/llavoie/fbe2106baed00746ec8e

On Thu, Jun 4, 2015 at 9:39 PM, Lauren Lavoie <notifications@github.com>
wrote:

> A few more questions:
>
>    - The sample data we have from the government web service (
>    https://gist.github.com/llavoie/fbe2106baed00746ec8e) is
>    tab-delimited, so don't we need to use CsvPreference.TAB_PREFERENCE here?
>    - The sample data also has a few junk lines at the top of the file --
>    will those be skipped?
>    - The sample data also has quite a few more columns than the ones that
>    this code is processing -- will the CsvImporter skip those or do they need
>    to be handled somewhere? e.g. do we need to include them in the
>    ProcessorMapping even if they don't get saved on the FLW record?
>
> —
> <https://github.com/motech-implementations/mim/pull/295#issuecomment-109158872>
> .
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

llavoie:
@frankhuster I just added the IT. It fails for now, so we can let @kamknitt take a look when he's back in the office.

frankhuster:
Nice!

On Fri, Jun 5, 2015 at 9:13 AM, Lauren Lavoie <notifications@github.com>
wrote:

> @frankhuster <https://github.com/frankhuster> I just added the IT. It
> fails for now, so we can let @kamknitt <https://github.com/kamknitt> take
> a look when he's back in the office.
>
> —
> <https://github.com/motech-implementations/mim/pull/295#issuecomment-109346774>
> .
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

larubbio:
We also have some changes to make to the datamodel.  I'm not sure where Abhishek was looking, but apparently the codes are not unique, but rather namespaced within the parent.  So district code is only unique within a state, and taluka code is only unique with a district etc.

So we'll have to change the data model to reflect the two part key, and the import to use code.  I'm not sure if we want to do that as part of this ticket, or a new one.

llavoie:
Yuck. I'll need to change how I'm doing Kilkari beneficiary imports too then.

larubbio:
Yep.  Also you'll notice the csv portion of the file doesn't have state in
it, so that will either have to be read from those first few header lines,
or added as a parameter to the http post.

-Rob

flw/src/main/java/org/motechproject/nms/flw/controller/FrontLineWorkerImportController.java
kamknitt: You are right. I just wanted to be consistent with other csv imports.

frankhuster:
Hey man, I'm going through the pull request list and see this has not seen any action in a long time.
Do you need any help? (highly unlikely, I imagine I'm probably the one needing help from you)

I chatted with @llavoie and she mentioned she fixed a similar error in her code, you might want to check it out. She also said she's happy to take over this pull request and fix it herself.

Thanks dude!

@kamknitt

PR #300, opened 2015-06-03
NMS.GEN.SECU.004

srivastavaAbhi:
No personal information like name,mobileno,dob,lmp are logged currently.
Though,its better to remove them from toString method which cuts down the possibility of such information to be logged substantially.

### DIFF
--- a/flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
@@ -3,6 +3,7 @@ package org.motechproject.nms.flw.domain;
 import org.joda.time.DateTime;
 import org.motechproject.mds.annotations.Entity;
 import org.motechproject.mds.annotations.Field;
+import org.motechproject.mds.domain.MdsEntity;
 import org.motechproject.nms.region.domain.District;
 import org.motechproject.nms.region.domain.LanguageLocation;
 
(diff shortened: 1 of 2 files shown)


motech-gerrit:

frankhuster:
ok to test

flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
```diff
@@ -177,8 +177,8 @@ public int hashCode() {
     public String toString() {
         return "FrontLineWorker{" +
                 "id=" + id +
-                ", contactNumber='" + contactNumber + '\'' +
-                ", name='" + name + '\'' +
+                ", contactNumber='" + "***" + '\'' +
+                ", name='" + "***" + '\'' +
```
larubbio: Instead of replacing with "***" I would just remove the lines.

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscriber.java
```diff
@@ -170,9 +170,9 @@ public int hashCode() {
     @Override
     public String toString() {
         return "Subscriber{" +
-                "callingNumber=" + callingNumber +
-                ", dateOfBirth=" + dateOfBirth +
-                ", lastMenstrualPeriod=" + lastMenstrualPeriod +
+                "callingNumber=" + "***" +
+                ", dateOfBirth=" + "***" +
+                ", lastMenstrualPeriod=" + "***" +
```
larubbio: Instead of replacing with "***" I would just remove those fields.  I would also log id so the record can be found in the database if we need to debug something related to it.

frankhuster:
@srivastavaAbhi, I wonder if there are places (especially the ones dealing with OBD and CDR) that may also expose phone numbers in logs. Can you confirm?

srivastavaAbhi:
I can't find any personal information in the logs.

frankhuster:
Great, thanks!

PR #301, opened 2015-06-04
added test cases NMS_FT_151, NMS_FT_152 in NMS.FT.151.152.

PriyaJain5:
Added test cases NMS_FT_151, NMS_FT_152


motech-gerrit:

frankhuster:
ok to test

testing/src/test/java/org/motechproject/nms/testing/it/imi/TargetFileServiceBundleIT.java
```diff
@@ -298,6 +304,101 @@ public void createLargeFile() {
...
+             BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
+            DigestInputStream dis = new DigestInputStream(is, md);
+            while ((line = reader.readLine()) != null) {
+                recordCount++;
+                contents.add(line.split(",")[7]); //column 8 is for weekId
+            }
+        }
+        String md5Checksum = new String(Hex.encodeHex(md.digest()));
+
+        assertEquals((int)tfn.getRecordCount(), recordCount);
```
frankhuster: Seems to me you're expecting only one row. Verifying that the TFN matches is good, but I think you should also be verifying that it is equal to the number of records you're expecting, which is 1. Right?

frankhuster:
@PriyaJain5, do you need help on this?

frankhuster:
ok to test

PR #327, opened 2015-06-05
Implemented test cases NMS_FT_86, 87, 88

kanchannms:


motech-gerrit:

koshalt:
ok to test

testing/src/test/java/org/motechproject/nms/testing/it/api/KilkariControllerBundleIT.java
```diff
@@ -783,4 +785,34 @@ public void verifyFT77()
                 expectedJsonPattern, ADMIN_USERNAME, ADMIN_PASSWORD));
     }
 
+    @Ignore
```
koshalt: Couple of questions - why is this marked ignore? Do we already have a jira ticket for the failures? Can you add it here, if that's true.

Also, it is usually good practice to have a test case per scenario to make that they are aren't cascading failures and it is easy from a failure to pinpoint a particular scenario than mixing them up

nmsyogesh:
A JIRA issue has been raised for tracking this at https://applab.atlassian.net/browse/NMS-186.

frankhuster:
please merge

srivastavaAbhi:
Tracked with branch NMS.FT.KILKARICONTROLLERIT.
So closing this request.

PR #335, opened 2015-06-09
Added test case nNMS_FT_200.

PriyaJain5:
This test is failing, raised bug in jira. BugID:https://applab.atlassian.net/browse/NMS-191. Added ignore to make sure build is not failing.


testing/src/test/java/org/motechproject/nms/testing/it/imi/ImiController_OBD_BundleIT.java
```diff
@@ -155,4 +156,28 @@ public void testCreateFileProcessedStatusRequestWithError() throws IOException,
...
+
+    /*
+    * Invoke "NotifyFileProcessedStatus" API having mandatory parameter
+    * fileProcessedStatus having invalid value(i.e status code which doesn’t exist in system).
+    */
+    @Test
+    //TODO :https://applab.atlassian.net/browse/NMS-191
+    @Ignore
+    public void verifyFT200() throws IOException, InterruptedException {
+        getLogger().debug("testCreateFileProcessedStatusRequestWithInvalidFileProcessedStatusError()");
```
frankhuster: Why is that here?

PR #337, opened 2015-06-09
Added test cases NMS_FT_201,202,203,204,205,206,207,208,209,210.

PriyaJain5:
Added 10 test cases from NMS_FT_201 to NMS_FT_210.


frankhuster:
lgtm, but looks like you're not testing invalid checksums, please add new test cases.

PR #339, opened 2015-06-09
Added test case NMS_FT_193.

PriyaJain5:
Mapped test cases NMS_FT_194 and NMS_FT_195 to NMS_FT_193.


frankhuster:
please merge

testing/src/test/java/org/motechproject/nms/testing/it/imi/ImiController_OBD_BundleIT.java
```diff
@@ -180,4 +180,28 @@ public void verifyFT200() throws IOException, InterruptedException {
...
     }
+
+    /*
+    * NMS_FT_193 :To check "NotifyFileProcessedStatus" API is rejected in case the file could not
+    * be copied or the file is not available.
+    * NMS_FT_194 : To check "NotifyFileProcessedStatus" API is rejected in case there is an error in checksum
+    * NMS_FT_195 : To check "NotifyFileProcessedStatus" API is rejected in case there is an error in records check.
+    */
+    @Test
+    public void verifyFT193_194_195() throws IOException, InterruptedException {
```
koshalt: Looks like this is only testing issue 193 and not 194 & 195?

PriyaJain5: yes, this test case is testing only 193, but I have mapped test cases 194 and 195 to this test case only as all the test cases are testing same piece of code except the enum value is different.

koshalt: I don't agree with this shortcut of assuming that they are all the same. The enums passed are explicitly different for each of those test cases. The internal behavior might be the same that you wouldn't know that in the black box scenario. 

I agree that we just raise the same alert in each of those sub-scenarios but you are not truly testing the different scenarios here but just the output of multiple scenarios which happen to have the same result.

@srivastavaAbhi lets chat about this when you are in.

srivastavaAbhi: Updated as per @koshalt comments.

koshalt:
looks good.

PR #343, opened 2015-06-09
Implemented NMS.FT.182 In which calling number greater than 10 digit …

PriyaJain5:
…should be valid.

Used Ignore as functionality is not available in current implementation.


frankhuster:
ok to test

PR #352, opened 2015-06-09
NMS.FT.153 verify initial state of subscription should be pending

PriyaJain5:


testing/src/test/java/org/motechproject/nms/testing/it/kilkari/SubscriptionServiceBundleIT.java
```diff
@@ -760,4 +762,22 @@ public void testDeleteOldCompletedSubscription() {
         subscriber = subscriberDataService.findByCallingNumber(2000000000L);
         assertEquals(1, subscriber.getSubscriptions().size());
     }
+    
+    @Ignore
```
koshalt: Please add a JIRA link here to the bug created, assuming that is the reason why this is marked as Ignore

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/SubscriptionServiceBundleIT.java
```diff
@@ -760,4 +762,22 @@ public void testDeleteOldCompletedSubscription() {
...
+    @Test
+    public void verifyFT153() {
+    	/*
```
koshalt: @AnkitGaurNms @PriyaJain5 @kanchannms @srivastavaAbhi Please make sure to add the comment for the method on top and not inside. I understand @frankhuster might have started this trend originally, but we should stick with the convention and move the comments to the top of the method.

frankhuster: My bad, sorry everyone & thanks for being chief stickler, @koshalt!

srivastavaAbhi: Hi Koshal,

Doesn't it look cleaner to have test case description ( pasted from test
case doc) just inside the method ?
It looked to me quite good when @Frank and I pasted it inside the method
body.

koshalt: @srivastavaAbhi This is more to do with conventions and keeping up with coding style of the rest of our code base. Method and test descriptions go on top of the method with internal comments being specific to a particular line(s).

PriyaJain5:
As this branch has been merged in SubscriptionServiceBundleIT branch

PR #374, opened 2015-06-10
Reduced the number failures to test maximum error threshold from 100 …

frankhuster:
…to 3

Just to save a little build time.


llavoie:
lgtm

PR #383, opened 2015-06-11
Composite keys for locations

frankhuster:
A location object code is unique amongst its sibling.
Two location objects can have the same code if they have different parent location objects.


frankhuster:
ok to test

llavoie:
Don't the various location *DataService lookups need to be updated as well? e.g. DistrictDataService.findByCode(code) can no longer be assumed to return a single District. So we'd need to change to findByCode(districtCode, stateCode) or similar.

frankhuster:
I'm working on that right now.

On Fri, Jun 12, 2015 at 10:03 AM, Lauren Lavoie <notifications@github.com>
wrote:

> Don't the various location *DataService lookups need to be updated as
> well? e.g. DistrictDataService.findByCode(code) can no longer be assumed to
> return a single District. So we'd need to change to
> findByCode(districtCode, stateCode) or similar.
>
> —
> <https://github.com/motech-implementations/mim/pull/383#issuecomment-111558441>
> .
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

frankhuster:
Talk about Pandora's box...

frankhuster:
No point building this just yet.

PR #384, opened 2015-06-12
Fixes #112 - deactivate pregnancy subscription

### LINKED ISSUE #112: NMS.GEN.MCTS.004 - MCTS Beneficiary (Kilkari subscriber) identification
A Kilkari Beneficiary shall be identified by a unique identifier that comprises of state specific MCTS Id along with State identifier. However, if the Kilkari subscriber activates their subscription by calling the IVR long codes (one for each pack) then they will remain anonymous unless and until MCTS data is uploaded where a pregnancy or birth record matches with the MSISDN number of the anonymous subscriber. For anonymous subscribers, they are only identified by their MSISDN + pack info.

It’s possible that a child MCTS record and pregnancy MCTS record could exist at the same time for the same woman. If the pregnancy record is present, then this record will be UPDATED if a new MCTS record is uploaded  that contains both a child MCTS ID and a pregnancy MCTS ID, where the pregnancy MCTS ID already exists in NMS. In other words, only one subscription will be maintained for the same woman, the pregnancy pack shall be de-activated (if activated) and child-pack shall be activated based on DOB of the child. If, when a MCTS child record is being uploaded, no pregnancy record is found then only the child-pack shall be activated based on DOB of the child. 

For details see Appendix A

llavoie:
Deactivate pregnancy subscription if a new child subscription is created via MCTS import with matching mother ID.

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/DeactivationReason.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/DeactivationReason.java
@@ -5,6 +5,7 @@ package org.motechproject.nms.kilkari.domain;
  */
 public enum DeactivationReason {
     DEACTIVATED_BY_USER,
+    LIVE_BIRTH,
     MISCARRIAGE_OR_ABORTION,
     STILL_BIRTH,
     CHILD_DEATH,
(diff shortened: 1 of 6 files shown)


koshalt:
Looks good. 

One follow up question - Is it possible for someone to get started with a child pack and then get enrolled in the pregnancy pack? Is this a scenario that we care about?

I'm trying to understand if it is possible to get updates from MCTS for a deactivated child record while the pregnancy is active? I tried reading through #112 ...but went cross-eyed :confounded:

llavoie:
I think it's theoretically allowed, at least I haven't seen it explicitly outlawed in the spec. This issue discusses rules for rejecting subscriptions if you want your head to hurt more: https://github.com/motech-implementations/mim/issues/159

-----Original Message-----

PR #385, opened 2015-06-12
NMS-181

srivastavaAbhi:
Bug fix for the KK deployment issue. FT case added as well.


api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -191,6 +192,20 @@ protected State getStateForFrontLineWorker(FrontLineWorker flw, Circle circle) {
...
+
+
+    protected State getStateFromCircle(Circle circle) {
+
+        State state = null;
+
+        if (circle != null) {
             List<State> states = circle.getStates();
 
             if (states.size() == 1) {
```
frankhuster: What happens if there is more than one state?

srivastavaAbhi: If there are more than 1 state , in that case,if kk is deployed in any of
these we need to handle it as deployed service.

I shall update the code after verification from SRS.

Thanks,
Abhishek

srivastavaAbhi: It looked to me we are not checking the deployment status and whitelist
status for the first time caller in NMS.

Raised a bug : https://applab.atlassian.net/browse/NMS-205

larubbio:
Is someone still looking at the PR?

frankhuster:
@larubbio I started, but got sidetracked. You're welcome to take it over, especially since you're more familiar with whitelisting/deployment.

srivastavaAbhi:
Two bugs can be resolved with this PR.
Please review it.

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
larubbio: Just to be clear, if there is more than one state do we consider it deployed if it is deployed in one, or not deployed?  The code seems to assume deployed and this thread doesn't seem to have a conclusive answer.

srivastavaAbhi: Clarified from Saurabh that we need not to do the validation check if we are not sure about the state.
Thus, I have closed that bug NMS-205 but updated the code accordingly.

testing/src/test/java/org/motechproject/nms/testing/it/api/UserControllerBundleIT.java
```diff
@@ -160,6 +160,16 @@ private void createKilkariTestData() {
                 sh.pregnancyPack(), SubscriptionOrigin.IVR);
     }
 
+    private void createKilkariUndeployTestData() {
```
larubbio: Since this is only used in a single method I might just inline it in that test.  Not a big deal and the change shouldn't be blocked on that.

larubbio:
lgtm

PR #386, opened 2015-06-15
Location entity codes are only valid along with the parent location entity

frankhuster:
Except for the top-level State entity.
To search for a HealthBlock, for example, you'll need to provide the parent Taluka in addition to the health block code.


region/src/main/java/org/motechproject/nms/region/domain/HealthBlock.java
```diff
@@ -38,11 +40,14 @@
...
 
     @Field
-    @Unique
     @Column(allowsNull = "false")
     @NotNull
     private Long code;
 
+    // used by CSV import as a placeholder for the taluka
+    @Ignore
+    private String talukaCode;
```
kamknitt: I would avoid adding this kind of fields to entity classes (looks like a code smell). If such a placeholder is really needed, consider creating an intermediate class or use a map to store entity properties.

region/src/main/java/org/motechproject/nms/region/domain/HealthFacility.java
```diff
@@ -40,6 +41,10 @@
     @NotNull
     private HealthFacilityType healthFacilityType;
 
+    // used by CSV import as a placeholder for the healthBlock
+    @Ignore
+    private long healthBlockCode;
```
kamknitt: As above

region/src/main/java/org/motechproject/nms/region/domain/Taluka.java
```diff
@@ -46,6 +49,10 @@
     @NotNull
     private District district;
 
+    // used by CSV import as a placeholder for the district
+    @Ignore
+    private String districtCode;
```
kamknitt: As above

region/src/main/java/org/motechproject/nms/region/domain/HealthSubFacility.java
```diff
@@ -25,11 +27,13 @@
     private String regionalName;
 
     @Field
-    @Unique
     @Column(allowsNull = "false")
     @NotNull
     private Long code;
 
+    @Ignore
+    private long healthFacilityCode;
```
kamknitt: As above

region/src/main/java/org/motechproject/nms/region/domain/Village.java
```diff
@@ -44,13 +45,21 @@
...
     @Field
+    @NotNull
+    @Min(0)
     // This is the village code for a non-census village.
-    private Long svid;
+    private long svid;
+
+    // used by CSV import as a placeholder for the taluka
+    @Ignore
+    private String talukaCode;
```
kamknitt: As above

region/src/main/java/org/motechproject/nms/region/service/impl/BaseVillageImportService.java
```diff
@@ -0,0 +1,43 @@
...
+        District district = (District) getParent(PARENT_DISTRICT);
+        if (district == null) {
+            throw new CsvImportException("No district provided!");
+        }
+
+        try {
+            taluka = talukaService.findByDistrictAndCode(district, village.getTalukaCode());
+        } catch (NumberFormatException e) {
+            throw new CsvImportDataException(String.format("Invalid taluka: %s",
+                    (String) village.getTalukaCode()), e);
```
kamknitt: Redundant cast.

region/src/main/java/org/motechproject/nms/region/service/CensusVillageImportService.java
```diff
@@ -1,4 +1,7 @@
 package org.motechproject.nms.region.service;
 
+import org.motechproject.nms.region.domain.District;
+
 public interface CensusVillageImportService extends LocationDataImportService {
+    void addParent(District district);
```
kamknitt: Aren't all those addParent(..) methods supposed to be called from LocationDataImportController too?

region/src/main/java/org/motechproject/nms/region/service/impl/BaseLocationImportService.java
```diff
@@ -13,17 +13,33 @@
...
 public abstract class BaseLocationImportService<T> {
 
+    public static final String PARENT_STATE = "state";
+    public static final String PARENT_DISTRICT = "district";
+    public static final String PARENT_TALUKA = "taluka";
+    public static final String PARENT_HEALTH_BLOCK = "healthBlock";
+
     private Class<T> type;
     private MotechDataService<T> dataService;
+    private Map<String, Object> parents;
```
kamknitt: Placing a property in a service makes the service stateful. We should avoid it, as it makes it prone to concurrency issues. Consider passing the map as a parameter to the importData(..) method.

kamknitt:
The most crucial part to fix is the 'parents' property in the BaseLocationImportService class.
I've refactored the CsvInstanceImporter a bit (and opened a PR: https://github.com/motech-implementations/mim/pull/389) and added the ability to use values from already procesed cells to determine the value of consecutive cells. I thought that maybe you will find it useful while working on this PR.

frankhuster:
Great feedback, thanks!
I'll get working on it.

region/src/main/java/org/motechproject/nms/region/service/impl/BaseVillageImportService.java
koshalt: removed cast.

frankhuster:
Now uses @kamknitt's nice solution for accessing sibling column values in `CsvInstanceImporter`.

kamknitt:
Looks good to me.

PR #389, opened 2015-06-16
CsvInstanceImporter improvement

kamknitt:
Added the ability to use values from already procesed cells to determine the value of consecutive cells.
Could be used (for example) to determine the village based on taluka.
To ensure the cell processors execution order, use the LinkedHashMap for processor mapping and add processors in the desired order.

Example: using the values of 'bool' and 'number' cells to determine the value of the 'string' cell value:

```java
private Map<String, CellProcessor> getProcessorMapping() {
    // note the linked hash map!
    Map<String, CellProcessor> mapping = new LinkedHashMap<>();
    final Store store = new Store();
    mapping.put("bool", store.store("stored_bool", new GetBoolean()));
    mapping.put("number", store.store("stored_number", new GetLong()));
    mapping.put("string", new GetString() {
        @Override
        public Object execute(Object value, CsvContext context) {
            return super.execute(value, context) + ", " +
                    "bool is: " + store.get("stored_bool") + ", " +
                    "number is: " + store.get("stored_number");
        }
    });
    return mapping;
}
```


csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -17,14 +30,124 @@ public CsvInstanceImporter(Class<T> type) {
 
     public T read() throws IOException {
         if (isOpen()) {
-            return getCsvReader().read(type, getFieldNames(), getProcessors());
+            T instance = createInstance();
+            Map<String, String> cells = csvReader.read(columnNames);
+            if (null != cells) {
```
frankhuster: We've moved away from doing `null != value` comparisons since java won't allow non bool assignments in if clauses 
and our checkstyle rules break anyway.

frankhuster: oh, and if you're going to do something on both sides of a comparison, it makes the code more readable to test for equality than to test for non equality, so the following `if` construct would me more readable:
```
if (cells == null) {
    return null;
} else {
    ....
}
```

csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -17,14 +30,124 @@ public CsvInstanceImporter(Class<T> type) {
...
+                return null;
+            }
         } else {
             throw new IllegalStateException("CsvImporter is closed");
         }
     }
 
+    public void open(Reader reader, CsvPreference preferences, Map<String, CellProcessor> processorMapping, Map<String, String> fieldNameMapping)
+            throws IOException {
+        if (null == this.csvReader) {
```
frankhuster: why use `this.csvReader` here and just `csvReader` two lines below?

csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -17,14 +30,124 @@ public CsvInstanceImporter(Class<T> type) {
...
+        } else {
+            throw new IllegalStateException("CsvImporter is already open");
+        }
+    }
+
     @Override
-    protected ICsvBeanReader createCsvReader(Reader reader, CsvPreference preferences) {
-        return new CsvBeanReader(reader, preferences);
+    public void close() throws IOException {
+        if (null != csvReader) {
```
frankhuster: why not use `isOpen()`?

csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -17,14 +30,124 @@ public CsvInstanceImporter(Class<T> type) {
...
+            csvReader.close();
+            csvReader = null;
+            columnNames = null;
+            fieldMapping = null;
+            processorMapping = null;
+        }
+    }
+
+    public int getLineNumber() {
+        if (null != csvReader) {
```
frankhuster: why not use `isOpen()`?

csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -17,14 +30,124 @@ public CsvInstanceImporter(Class<T> type) {
...
+            return csvReader.getLineNumber();
+        } else {
+            return -1;
+        }
+    }
+
+    public int getRowNumber() {
+        if (null != csvReader) {
```
frankhuster: why not use `isOpen()`?

csv/src/main/java/org/motechproject/nms/csv/utils/CsvMapImporter.java
```diff
@@ -1,25 +1,84 @@
...
+        } else {
+            throw new IllegalStateException("CsvImporter is already open");
+        }
+    }
+
     @Override
-    protected ICsvMapReader createCsvReader(Reader reader, CsvPreference preferences) {
-        return new CsvMapReader(reader, preferences);
+    public void close() throws IOException {
+        if (null != csvReader) {
```
frankhuster: why not use `isOpen()`?

kamknitt:
Thanks for comments and fixes!

PR #400, opened 2015-06-18
Fixes #399

### LINKED ISSUE #399: KK - Never need to check for whitelisting
From Pankaj/Saurabh:
>for Kilkari there will not be whitelisting services.

frankhuster:


llavoie:
Should this be closed?

llavoie:
Hey @larubbio - yeah, we agreed that this was not the right fix, that's why we didn't merge it. There was a mail thread with Saurabh that we all had different interpretations of...

larubbio:
@llavoie Then can we close #399?

llavoie:
I'm not sure. I'll dig up the mail thread and forward to you -- depending on how you interpret it we can decide whether there's any work to do.

PR #401, opened 2015-06-19
Better checksum & better OBD file naming

frankhuster:


frankhuster:
@koshalt, wanna CR this when it's built?

imi/src/main/java/org/motechproject/nms/imi/service/impl/ChecksumHelper.java
```diff
@@ -0,0 +1,17 @@
...
+import java.io.FileInputStream;
+import java.io.IOException;
+
+
+public final class ChecksumHelper {
+
+    private ChecksumHelper() { }
+
+    public static String checksum(File file) throws IOException {
+            return DigestUtils.sha1Hex(new FileInputStream(file));
```
koshalt: Nothing wrong here... but if this is just a 1 line return, can we just replace the call line to this function with the return line here? :thought_balloon:

koshalt:
lgtm. feel free to merge if you don't think the comment is actionable right now. more of a nitpick than anything else

PR #407, opened 2015-06-22
Created csv package in org/motechproject/nms/region/service/impl

frankhuster:


larubbio:
There is an import service in kilkari (and maybe other modules) as well.  Do you want to move that one in another commit?

llavoie:
I'll move it -- I'm working on the update cases right now and have some refactoring in progress.

-----Original Message-----

PR #414, opened 2015-06-23
Fixes NIP-38

frankhuster:


imi/src/main/java/org/motechproject/nms/imi/service/impl/ChecksumHelper.java
```diff
@@ -12,6 +12,6 @@
     private ChecksumHelper() { }
 
     public static String checksum(File file) throws IOException {
-            return DigestUtils.sha1Hex(new FileInputStream(file));
+            return DigestUtils.md5Hex(new FileInputStream(file));
```
koshalt: md5 :laughing: lol ... so h4x0r3d

koshalt:
lgtm

PR #417, opened 2015-06-24
Fix for invalid module assigned to entities

tstalka:

PR #422, opened 2015-06-24
Resubmitting tstalka's fix for incorrect module assigned to entities

llavoie:
Re-doing PR #417 so that it may be tested via our CI


larubbio:
lgtm

PR #425, opened 2015-06-25
NMS-182

srivastavaAbhi:
SaveCall detail may have only 1 element in Array.
Check has been updated and TC updated as per latest understanding.

JIRA issue : https://applab.atlassian.net/browse/NMS-182

### DIFF
--- a/testing/src/test/java/org/motechproject/nms/testing/it/api/KilkariControllerBundleIT.java
+++ b/testing/src/test/java/org/motechproject/nms/testing/it/api/KilkariControllerBundleIT.java
@@ -1274,6 +1274,8 @@ public class KilkariControllerBundleIT extends BasePaxIT {
     /**
      * To verify the that Save Inbox call Details API request should not succeed with content provided for only 1
      * subscription pack and not even a place holder for second Pack details
+     *
+     * //TODO : FT case to be modified .Refer NMS -182
      */
     @Test
     public void verifyFT35() throws IOException, InterruptedException {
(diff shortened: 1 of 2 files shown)


testing/src/test/java/org/motechproject/nms/testing/it/api/KilkariControllerBundleIT.java
```diff
@@ -1274,6 +1274,8 @@ public void verifyFT85() throws IOException, InterruptedException {
     /**
      * To verify the that Save Inbox call Details API request should not succeed with content provided for only 1
      * subscription pack and not even a place holder for second Pack details
+     *
+     * //TODO : FT case to be modified .Refer NMS -182
```
frankhuster: Why not fix this as part of this pull request?

srivastavaAbhi: Code is modified, only thing left is Ft doc

Thanks,
Abhishek

frankhuster: OK, so I assume you will change the comment for this method (which currently doe snot match the code) when the FT description is updated?

srivastavaAbhi: FT description to be updated in next version of FT doc.
I have conveyed this to Kanchan.

Issue #377, opened 2015-06-11
A history/change table for FLW language updates


larubbio:
A history/change table shall be created with old and new language for the existing FLW when language is updated.

Pulled out of #124

larubbio:
Looking at the code this looks complete

Issue #137, opened 2015-04-25
NMS.GEN.LOG.005 - FLW and Kilkari beneficiary history tracking


larubbio:
NMS system shall maintain a history of all changes made to FLW and MCTS-beneficiary records via code.  Tracking of changes via UI is tracked under #475

llavoie:
turn on auditing

larubbio:
Instead of turning on auditing I was thinking of creating a history tracking service.  When a field is changed via a setter the history tracking service can be called to record the old and new value.  We can then use the @InstanceLifecycleListener(InstanceLifecycleListenerType.PRE_UPDATE) to flush the history tracking object.

larubbio:
@llavoie What object would you consider the MCTS beneficiary?  The Subscriber, MctsBeneficiary, MctsMother or MctsChild?

llavoie:
MctsMother and MctsChild would both be considered beneficiaries (both
extend MctsBeneficiary).

On Wed, Jun 24, 2015 at 3:41 PM, Rob LaRubbio <notifications@github.com>
wrote:

> @llavoie <https://github.com/llavoie> What object would you consider the
> MCTS beneficiary? The Subscriber, MctsBeneficiary, MctsMother or MctsChild?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/motech-implementations/mim/issues/137#issuecomment-115034491>
> .
>



-- 
Lauren Lavoie
Software Development Manager, Mobile Health Innovations
Grameen Foundation
2101 4th Ave | Suite 1550 | Seattle, WA 98121
206-325-6690 ext 254
llavoie@grameenfoundation.org | Skype:laurenlavoie
Facebook <http://www.facebook.com/StopPovertyNow>│Twitter
<http://twitter.com/grameenfdn>│YouTube
<http://www.youtube.com/user/grameen?blend=6&ob=5> | www.motechsuite.org
*Connecting the World's Poor to Their Potential*

larubbio:
If i have a history tracking table, could I just relate
to nms_mcts_beneficiaries or would I need to relate to the other tables?

llavoie:
I'm not sure -- we'd need to test it out.

On Wed, Jun 24, 2015 at 3:43 PM, Rob LaRubbio <notifications@github.com>
wrote:

> If i have a history tracking table, could I just relate
> to nms_mcts_beneficiaries or would I need to relate to the other tables?
>
[...]

PR #427, opened 2015-06-25
#377: A history/change table for FLW language updates

kamknitt:


larubbio:
There is another ticket for recording all changes to flws and beneficiaries.  I think we should roll this work into that ticket. #137

larubbio:
@kamknitt Can you close this pull request and look #137 Ideally for that we'd have single table with a string description of the changes in a parseable format.  ie: field(old, new)

so something like language(HIN, ORI), msisdn(2323123, 34534534) etc ...

Additionally the table should have a foreign key to flw, mcts_child and mcts_mother and maybe an enum for the change type (ADD, DELETE, UPDATE).

In a perfect world we'd be able to mark the entities we want to track with an annotation and then have it watch all setters.  In the event a field is set twice it should only be listed once with the original value and the final value (so the intermediate value would be ignored).  

If we have a global user that we can store in the table that would be good as well.

Issue #151, opened 2015-04-26
NMS.MA.COURSE.004 - Maintain count of course attempts


larubbio:
MA service shall maintain count of number of times the course is attempted and also allow the user to reattempt the chapter or restart the whole course from beginning.

There shall be no limitation on number of re-attempts.

koshalt:
The count for the number of times the course is attempted is captured by means of CallDetails where we set the service name.

Every MA service call details request represents 1 access of the course and this requirement is then fulfilled at the course level or at an individual FLW level by writing the appropriate ETL job to filter by service and/or group by FLW id.

The service also does not prevent a user from re-attempting a chapter or a sub-unit since the flow for it is controlled at the ivr level. We define the course structure and let IVR walk through it with the duplicate and non-skippable guarantees. Nothing else to do on the motech side of things.

Marking as resolved.

llavoie:
We realized that our earlier interpretation of this requirement was incorrect -- reopening. @koshalt will fix this based on an existing field in mTraining.

koshalt:
fixed and checked in. The data will be tracked in activity records table of mtraining when a user completes a course (regardless of pass or fail)

PR #429, opened 2015-06-25
Adding a way to track completion count for MA

koshalt:
Helps track the course completion count for a user.

### DIFF
--- a/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
+++ b/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
@@ -1,6 +1,7 @@
 package org.motechproject.nms.mobileacademy.service.impl;
 
 import org.apache.commons.io.IOUtils;
+import org.joda.time.DateTime;
 import org.json.JSONObject;
 import org.motechproject.alerts.contract.AlertService;
 import org.motechproject.alerts.domain.AlertStatus;
@@ -9,7 +10,10 @@ import org.motechproject.config.core.constants.ConfigurationConstants;
 import org.motechproject.event.MotechEvent;
 import org.motechproject.event.listener.EventRelay;
 import org.motechproject.event.listener.annotations.MotechListener;
+import org.motechproject.mtraining.domain.ActivityRecord;
+import org.motechproject.mtraining.domain.ActivityState;
 import org.motechproject.mtraining.domain.Bookmark;
+import org.motechproject.mtraining.service.ActivityService;
 import org.motechproject.mtraining.service.BookmarkService;
 import org.motechproject.nms.mobileacademy.domain.CompletionRecord;
 import org.motechproject.nms.mobileacademy.domain.NmsCourse;
@@ -60,6 +64,11 @@ public class MobileAcademyServiceImpl implements MobileAcademyService {
     private BookmarkService bookmarkService;
 
     /**
+     * Activity service to track user completion data
+     */
+    private ActivityService activityService;
+
+    /**
      * Completion record data service
      */
     private CompletionRecordDataService completionRecordDataService;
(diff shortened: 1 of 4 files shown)


koshalt:
Fixes #151

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -183,6 +193,9 @@ public void setBookmark(MaBookmark saveBookmark) {
                 && saveBookmark.getScoresByChapter().size() == CHAPTER_COUNT) {
 
             LOGGER.debug("Found last bookmark and 11 scores. Starting evaluation & notification");
+            // Create an activity record here since pass/fail counts as 1 try
+            activityService.createActivity(
+                    new ActivityRecord(callingNumber.toString(), null, null, null, null, DateTime.now(), ActivityState.COMPLETED));
```
larubbio: I wish java had named parameters so instead of (null, null, null) you could do something like (name=null, somethingElse=null ...)  That or default values.

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
```diff
@@ -64,6 +68,8 @@
     public void setupMobileAcademy() {
 
         completionRecordDataService.deleteAll();
+        activityDataService.deleteAll();
+        bookmarkDataService.deleteAll();
```
larubbio: Should these go here on in the testingService.clearDatabase()

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
```diff
@@ -198,8 +205,30 @@ public void testSetLastBookmark() {
...
+        bookmark.setBookmark("Chapter11_Quiz");
+        Map<String, Integer> scores = new HashMap<>();
+        for (int i = 1; i < 12; i++) {
+            scores.put(String.valueOf(i), ((int) (Math.random() * 100)) % 5);
+        }
+        bookmark.setScoresByChapter(scores);
+        maService.setBookmark(bookmark);
+        int completionCountAfter = activityDataService.findRecordsForUser(String.valueOf(callingNumber)).size();
+
+        assertEquals(completionCountBefore + 1, completionCountAfter);
```
larubbio: If you are cleaning the database shouldn't you now the value?  How about assert the completionCountBefore == 0 around line 212 and then assert that it is 1 here?

koshalt: Yes, but I would also be assuming that the earlier call to the db succeeded (which it should have).. but in the event that the clear failed, this would still pass?

larubbio:
In general it looks fine, but it's unclear to me how this completes the requirement in a way different from just tracking the call details.  Is setBookmark called multiple times for a call?  Also @llavoie mentioned on the ticket that our understanding of the req has changed.  How has it changed?

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
larubbio: If you assert completionCountBefore == 0 around line 212 then are you assuming it succeeded or verifying it succeeded?  (I hope if it fails it raises an exception and doesn't return 0 :) 

I think if the clear fails we don't want the test passing.

koshalt:
Our earlier understanding was exactly what you mentioned in terms of # of calls made for the course attempt, but this changed to  (derived from saurabh's recent emails)  - 
- where they just treated the completion count as the number of times a user ran through the entire extent of the course.. regardless of whether they passed or failed (also regardless of how many phone calls or call details were sent)

This satisfies that requirement  by - 
- only adding an activity record every time the very last bookmark is set, at which point we do the evaluation and sms notification if needed.

larubbio:
lgtm

PR #431, opened 2015-06-26
#141 - IVR Interaction logging w/o PII

### LINKED ISSUE #141: NMS.GEN.LOG.009 - IVR Interaction logging
NMS system shall log all interactions with IVR including but not limited to the following:
- Incoming Call 
- Outgoing Call 
- Outgoing SMS 
- Bookmark information
- Other Message Exchanges

frankhuster:
Fixes #141

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/BaseController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/BaseController.java
@@ -1,5 +1,6 @@
 package org.motechproject.nms.api.web;
 
+import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.motechproject.nms.api.web.contract.BadRequest;
 import org.motechproject.nms.api.web.exception.NotAuthorizedException;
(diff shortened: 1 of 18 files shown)


api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -63,6 +66,14 @@
     private FrontLineWorkerService frontLineWorkerService;
 
 
+    protected static void log(final String endpoint, final String s) {
+        LOGGER.debug(IVR_INTERACTION_LOG.format(endpoint) + (StringUtils.isBlank(s) ? "" : " : " + s));
```
larubbio: I think instead of debug we might want to log this at INFO

api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
```diff
@@ -54,6 +55,9 @@
     @Transactional
     public void saveCallDetails(@PathVariable String serviceName,
                                 @RequestBody CallDetailRecordRequest callDetailRecordRequest) {
+
+        log(String.format("/%s/callDetails", serviceName), LogHelper.nullOrString(callDetailRecordRequest));
```
larubbio: Not a huge deal, but Spring MVC should have an api to get the URI

http://stackoverflow.com/questions/3686808/spring-3-requestmapping-get-path-value

larubbio:
lgtm

PR #433, opened 2015-06-29
NMS-211

srivastavaAbhi:
Fix NMS-211,NMS-210 bug.
Also fixed NPE coming in flow of multiple state in same circle.

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
@@ -255,7 +255,10 @@ public class KilkariController extends BaseController {
             throw new NotFoundException(String.format(NOT_FOUND, "languageLocationCode"));
         }
 
+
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
```diff
@@ -272,15 +275,22 @@ public void createSubscription(@RequestBody SubscriptionRequest subscriptionRequ
...
+
+
+
         Set<State> stateSet = languageService.getAllStatesForLanguage(language);
 
         if (stateSet.size() == 1) {
             return stateSet.iterator().next();
         }
 
+        if (null != circle) {
```
frankhuster: We've agreed to use the opposite notation for these:
`if (circle != null)`

srivastavaAbhi: Incorporated.

PR #434, opened 2015-06-29
NIP-47

srivastavaAbhi:
Changed the code as per API document, status code 202 is checked for success response.


frankhuster:
Looks good, any way we can have an IT for that?

PR #439, opened 2015-06-30
Nip 51

srivastavaAbhi:
Epoch timestamp sent by IVR is in Seconds.
Thus, I have updated a method to convert the value of timestamp in millis for java api to convert it into DateTime format.


larubbio:
lgtm

Issue #142, opened 2015-04-25
NMS.GEN.LOG.010 - Audit trail logging


larubbio:
NMS system shall have audit trail logs that includes but not limited to the following:
- Any NMS CSV upload 

The audit trail log shall include event details, date/time of event and user credential information among others.

frankhuster:
Does "Audit trail log" mean an entry in a database table or a line in the tomcat log file or something else?

koshalt:
Pretty sure this is db. Tomcat logs are usually ephemeral (and for debugging)

On a side note, we were originally going to rely on history tracking to get this done, from what I remember.

larubbio:
We were going to rely on history tracking, but I'm concerned about how much data it might end up saving.

For the CSV uploads I think we need a seperate audit table, hopefully we can add that in one place in the csv module.

For changes through the UI that might be difficult since the only way we can hook into it would be with a trigger or the history tracking.

For changes to business logic I think we can turn on history tracking (actually maybe not.  my concern with history tracking is that all related objects also have to have history tracking turned on.  Then for each change the entire graph of objects are copied)  So tracking history for config objects is fine, but for things with relationships to domain tables it might get messy

PR #442, opened 2015-06-30
Partial fix for #142

frankhuster:
Records all CSV insert/update interactions in the database.

### DIFF
--- a/csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
+++ b/csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
@@ -28,8 +28,8 @@ public class CsvAuditServiceImpl implements CsvAuditService {
 
         if (outcome.length() > CsvAuditRecord.MAX_OUTCOME_LENGTH) {
             truncatedOutcome = outcome.substring(0, CsvAuditRecord.MAX_OUTCOME_LENGTH);
-            LOGGER.warn("The provided outcome field was more than {} characters and was truncated",
-                    CsvAuditRecord.MAX_OUTCOME_LENGTH - FAILURE.length());
+            LOGGER.warn("The provided outcome field was more than {} characters and was truncated, original " +
+                            "outcome: {}", CsvAuditRecord.MAX_OUTCOME_LENGTH - FAILURE.length(), outcome);
         } else {
             truncatedOutcome = outcome;
         }
--- a/csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
+++ b/csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
@@ -1,0 +1,47 @@
+package org.motechproject.nms.csv.service.impl;
+
+import org.motechproject.nms.csv.domain.CsvAuditRecord;
+import org.motechproject.nms.csv.repository.CsvAuditRecordDataService;
+import org.motechproject.nms.csv.service.CsvAuditService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+
+@Service("csvAuditService")
+public class CsvAuditServiceImpl implements CsvAuditService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CsvAuditServiceImpl.class);
+    public static final String SUCCESS = "Success";
+    public static final String FAILURE = "Failure: ";
+
+    private CsvAuditRecordDataService csvAuditRecordDataService;
+
+    @Autowired
+    public void setCsvAuditRecordDataService(CsvAuditRecordDataService csvAuditRecordDataService) {
+        this.csvAuditRecordDataService = csvAuditRecordDataService;
+    }
+
+    private void audit(String file, String endpoint, String outcome) {
+        String truncatedOutcome;
+
+        if (outcome.length() > CsvAuditRecord.MAX_OUTCOME_LENGTH) {
+            truncatedOutcome = outcome.substring(0, CsvAuditRecord.MAX_OUTCOME_LENGTH);
+            LOGGER.warn("The provided outcome field was more than {} characters and was truncated",
+                    CsvAuditRecord.MAX_OUTCOME_LENGTH - FAILURE.length());
+        } else {
+            truncatedOutcome = outcome;
+        }
+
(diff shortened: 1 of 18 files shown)


csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    public void setCsvAuditRecordDataService(CsvAuditRecordDataService csvAuditRecordDataService) {
+        this.csvAuditRecordDataService = csvAuditRecordDataService;
+    }
+
+    private void audit(String file, String endpoint, String outcome) {
+        String truncatedOutcome;
+
+        if (outcome.length() > CsvAuditRecord.MAX_OUTCOME_LENGTH) {
+            truncatedOutcome = outcome.substring(0, CsvAuditRecord.MAX_OUTCOME_LENGTH);
+            LOGGER.warn("The provided outcome field was more than {} characters and was truncated",
```
larubbio: Can you also log the full outcome here?

flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
```diff
@@ -69,11 +75,14 @@ public void importFrontLineWorkers(@RequestParam MultipartFile csvFile) {
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 frontLineWorkerImportService.importData(new InputStreamReader(in));
+                csvAuditService.auditSuccess(csvFile.getName(), "/flw/import");
```
larubbio: Should this call move into the logError method?  I may have added some catch statements with my most recent PR so moving this into there would insure that the audit log is still written (since I assume I'm calling logError)

kilkari/src/main/java/org/motechproject/nms/kilkari/web/MctsBeneficiaryImportController.java
```diff
@@ -40,11 +54,14 @@ public void importMotherData(@RequestParam MultipartFile csvFile) {
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 mctsBeneficiaryImportService.importMotherData(new InputStreamReader(in));
+                csvAuditService.auditSuccess(csvFile.getName(), "/kilkari/mother/import");
```
larubbio: Same comment about moving into logError

kilkari/src/main/java/org/motechproject/nms/kilkari/web/MctsBeneficiaryImportController.java
```diff
@@ -57,11 +74,14 @@ public void importChildData(@RequestParam MultipartFile csvFile) {
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 mctsBeneficiaryImportService.importChildData(new InputStreamReader(in));
+                csvAuditService.auditSuccess(csvFile.getName(), "/kilkari/child/import");
             }
         } catch (CsvImportException e) {
+            csvAuditService.auditFailure(csvFile.getName(), "/kilkari/child/import", e.getMessage());
```
larubbio: Same comment about moving into logError

region/src/main/java/org/motechproject/nms/region/controller/LanguageLocationImportController.java
```diff
@@ -34,11 +36,14 @@ public void importLanguageLocationCodes(@RequestParam MultipartFile csvFile) {
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 languageLocationImportService.importData(new InputStreamReader(in));
+                csvAuditService.auditSuccess(csvFile.getName(), "/region/languageLocationCode/import");
             }
         } catch (CsvImportException e) {
+            csvAuditService.auditFailure(csvFile.getName(), "/region/languageLocationCode/import", e.getMessage());
```
larubbio: Same comment about moving into logError

region/src/main/java/org/motechproject/nms/region/controller/LocationDataImportController.java
```diff
@@ -44,25 +45,32 @@
...
+                    csvAuditService.auditSuccess(csvFile.getName(), endpoint);
                 } else {
-                    throw new IllegalArgumentException(String.format("Location type '%s' not supported", location));
+                    String error = String.format("Location type '%s' not supported", location);
+                    csvAuditService.auditFailure(csvFile.getName(), endpoint, error);
+                    throw new IllegalArgumentException(error);
                 }
             }
         } catch (CsvImportException e) {
+            csvAuditService.auditFailure(csvFile.getName(), endpoint, e.getMessage());
```
larubbio: Same comment about moving into logError

larubbio:
Small comments, but still lgtm.  Waiting for build to complete.

flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
frankhuster: It could, but you'd need to change logError to pass the file name.

csv/src/main/java/org/motechproject/nms/csv/service/impl/CsvAuditServiceImpl.java
frankhuster: Not sure I understand what you mean

larubbio: If the outcome exceeds 1000 characters, you write a log message saying it was too long, truncate it and put the shortened string in the db.  Why not also right the full string to the log.

frankhuster: Oh, I do understand. Yep, I'll do that. & the logError stuff too. And then I'll merge, ok?

larubbio: Sounds good.

llavoie:
@larubbio @frankhuster 
I think I have a different interpretation of this ticket than you did (not saying that I'm right :smile:, but would be good to discuss). I was expecting that in the CSV import code we would catch all exceptions and log errors at the granularity of each line of the CSV file (rather than a global audit entry per CSV file), and keep processing the file. Is that incorrect? For example, if I'm updating an MCTS beneficiary's location via the update CSV file, and there's a bad district ID for one entry, I would expect to log an error to the CSV audit table, skip that record, and move on. Is that not what's desired?

PR #443, opened 2015-07-01
Fixes NMS-213

larubbio:
Track the column number so it can be printed in the error message

### DIFF
--- a/csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
+++ b/csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
@@ -85,6 +85,20 @@ public class CsvInstanceImporter<T> implements Closeable {
         }
     }
 
+    public int getColumnNumber(String columnName) {
+        int columnNumber = 0;
+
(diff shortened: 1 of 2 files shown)


csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
```diff
@@ -85,6 +85,20 @@ public int getRowNumber() {
         }
     }
 
+    public int getColumnNumber(String columnName) {
+        int columnNumber = 0;
+
+        for (String column : columnNames) {
+            columnNumber += 1;
+
+            if (column.equals(columnName)) {
```
frankhuster: case sensitive?

frankhuster:
lgtm

csv/src/main/java/org/motechproject/nms/csv/utils/CsvInstanceImporter.java
larubbio: The are both from the same source so it shouldn't matter.

PR #445, opened 2015-07-01
NMS-188, NMS-185

larubbio:
Renable tests and in some cases update the expected result since we also return an error for the parent field.

I may have included changes for NMS-208 in this work as well due to sloppy branch management.  oops


frankhuster:
Always good to be removing @Ignore

PR #446, opened 2015-07-01
NMS-215

srivastavaAbhi:
Reset welcome message flag in processCallSummaryRecord.
Added FT case as well.

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
@@ -122,6 +122,8 @@ public class CsrServiceImpl implements CsrService {
     private void completeSubscriptionIfNeeded(Subscription subscription, CallSummaryRecord record,
                                               CallRetry callRetry) {
 
+        resetWelcomeFlagInSubscription(subscription);
+
         if (!subscription.isLastPackMessage(record.getContentFileName())) {
             // This subscription has not completed, do nothing
             return;
@@ -175,6 +177,7 @@ public class CsrServiceImpl implements CsrService {
             return;
         }
 
+        
         // We've already rescheduled this call, let's see if it needs to be re-rescheduled
 
         if ((subscription.getSubscriptionPack().retryCount() == 1) ||
@@ -190,6 +193,7 @@ public class CsrServiceImpl implements CsrService {
                 subscriptionDataService.update(subscription);
             }
 
+            resetWelcomeFlagInSubscription(subscription);
 
             callRetryDataService.delete(callRetry);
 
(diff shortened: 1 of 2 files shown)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -281,6 +281,7 @@ public void processCallSummaryRecord(MotechEvent event) {
 
         CallRetry callRetry = callRetryDataService.findBySubscriptionId(subscriptionId);
         Subscription subscription = subscriptionDataService.findBySubscriptionId(subscriptionId);
+        resetWelcomeFlagInSubscription(subscription);
```
larubbio: Do we want to reset the welcome flag in all cases or just success?

srivastavaAbhi: Discussed with @frankhuster .
We agreed to reset in all cases.

But now i am not sure if call drops in between welcome message, so are we able to play welcome message again or not ?

srivastavaAbhi: Discussed with @llavoie  just now.

We will reset it in case of success call and the calls for which we have tried maximum number of retries.

Is that sound ok ?

larubbio: sounds good to me.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -294,4 +299,12 @@ public void processCallSummaryRecord(MotechEvent event) {
             deactivateSubscription(subscription, callRetry);
         }
     }
+
+    private void resetWelcomeFlagInSubscription(Subscription subscription) {
+
+        if (subscription.getNeedsWelcomeMessage() == true) {
```
llavoie: nitpick: the check on line 305 isn't really needed

srivastavaAbhi: We can at least save a DB call with just one check for each KK subscription created via IVR.

srivastavaAbhi: method call on the line 196 can be removed.

llavoie: ok, once you remove the call in line 196 we can merge

srivastavaAbhi: Removed

PR #447, opened 2015-07-01
Fixes NMS-207

larubbio:
Return the subscription from processSubscriptionForBeneficiary in all cases.

cc @llavoie


llavoie:
looks good

PR #460, opened 2015-07-07
Removing Jira IDs for FTs that were already un-ignored

llavoie:


larubbio:
I was thinking we would leave those there as a record of the issue and discussion around it.

llavoie:
I see -- I don't feel strongly about this, but it looks like the Jira IDs for other test cases had been removed so I was just trying to be consistent.

PR #464, opened 2015-07-08
Fixes #454 - deactivate subscription due to maternal/child death

### LINKED ISSUE #454: Deactivate subscriptions if MCTS indicates maternal/child death
Ticket #116 indicates that subscriptions should be deactivated based on certain conditions reported through MCTS, including maternal/child death, which is indicated in the Entry_Type column as described below. However, MCTS does not seem to provide us that field at the time being. This ticket tracks the fact that this part of #116 is not done and awaits clarification from BBC and/or the MoH. Marking it as blocked. The other cases mentioned in #116 (abortion/miscarriage and stillbirth) are being implemented as described in that ticket.

- Death (Maternal /Child): The status will be available in the Entry_Type and the integer values in this field could be
o 1 - Active
o 2 - Migrated In
o 3 - Migrated Out
o 9 – Death.

llavoie:


larubbio:
lgtm.  Once build is cleaned up it can be merged.

llavoie:
I'm trying to debug this, but of course now I can't get it to fail when I run it in the debugger. ;) So this appears to be a DetachedFieldException in SubscriberDataService.update (generated code) on getting or setting the Subscriber.MctsChild's or Subscriber.MctsMother's state field, which is derived from Rob's new FullLocation interface.

Any insights?

larubbio:
I remember that when I was implementing that I first used the .equals methods on the objects. 

I then realized that they were not implemented correctly (since they compare the id, but not the parent and ids are only unique within a parent).

So I then modified all of the equals() methods to check the parent (and parent's parent) as necessary.  My tests then failed with detachedFieldExceptions.

So I then undid that change and went with the current implementation.

I notice the code creates the test data two different ways:

```java
        child.setState(stateDataService.findByCode(21L));
        child.setDistrict(child.getState().getDistricts().get(0));
```

```java
        child.setState(stateDataService.findByCode(21L));
        child.setDistrict(stateDataService.findByCode(21L).getDistricts().get(0));
```

Would it make a difference to instead do it this way:

```java
        child.setState(stateDataService.findByCode(21L));
        child.setDistrict(districtService.findByStateAndCode(child.getState() , 2L));
```

llavoie:
I'll try it when I get back home from my appointment. Thanks!

-----Original Message-----

llavoie:
This is either fixed now or I just had a lucky build. Ok to merge?

larubbio:
yes, do it quickly before it fails again. :)

PR #466, opened 2015-07-08
Fixes NIP-60 (really)

larubbio:
If the user needs the welcome message, set that as the filename.

@llavoie This feels wrong to me.  First, the ticket is hard to follow.  Once it was decided that the actual issue was something different then the description should be updated.  As far as I can tell this "fixes" it.  Although it seems odd that the user will call and only hear the week 1/welcome message when the actual message is week 5 or something else.  Shouldn't we return two messages, the week 1/welcome and the actual message?

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/domain/Subscription.java
@@ -22,7 +22,7 @@ public class Subscription {
     private static final int WEEKDAY2 = 1;
     private static final int WEEKDAY5 = 4;
     private static final int WEEKDAY6 = 5;
-
+    private static final int WEEKDAY7 = 6;
 
     @Field
     @Unique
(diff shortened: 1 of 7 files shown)


llavoie:
Yeah, I disagree with the spec for this. But also, as far as I can tell, this fix won't actually do what is being asked. We're supposed to play the welcome message even if the user has already heard it through IVR (i.e. needsWelcomeMessage would be false in this case), as long as it's the user's first week subscribed to the service.

larubbio:
Ok, do you want to take it over?

llavoie:
The reason I haven't taken it yet is that I haven't thought up a good fix. :) But I'll stew on it.

larubbio:
I can keep it if you want.

llavoie:
Presumably we could add a field for the "activationDate" (i.e. the date the subscription was activated) in addition to the startDate (i.e. the date corresponding to "week 1" of the subscription pack messages) -- and compare these when handling inbox requests. Unless you have an cleverer ideas. I can do this or you -- don't care.

larubbio:
Why don't you take it.  I'll look at another ticket.

llavoie:
CR please

larubbio:
lgtm

PR #468, opened 2015-07-09
Fixes NIP-56

frankhuster:
`generateTargetFile` now uses S1 and S2 as `ImiServiceId` field
for MCTS and IVR subscriptions, respectively

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/ImiServiceId.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/ImiServiceId.java
@@ -1,0 +1,25 @@
+package org.motechproject.nms.imi.service.impl;
+
+/**
+ * A method for IMI to treat OBD request in different ways
+ * see https://applab.atlassian.net/browse/NIP-56
+ */
+public enum ImiServiceId {
+    CHECK_DND,
+    DO_NOT_CHECK_DND;
+
+    public static final String CHECK_DND_STRINGVAL = "S1";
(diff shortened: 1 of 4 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/ImiServiceId.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * A method for IMI to treat OBD request in different ways
+ * see https://applab.atlassian.net/browse/NIP-56
+ */
+public enum ImiServiceId {
+    CHECK_DND,
+    DO_NOT_CHECK_DND;
+
+    public static final String CHECK_DND_STRINGVAL = "S1";
+    public static final String NO_NOT_CHECK_DND_STRINGVAL = "S2";
```
koshalt: is this a double negative? ergo, check dnd? :smile:

koshalt:
lgtm

imi/src/main/java/org/motechproject/nms/imi/service/impl/ImiServiceId.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A method for IMI to treat OBD request in different ways
+ * see https://applab.atlassian.net/browse/NIP-56
+ */
+public enum ImiServiceId {
+    CHECK_DND,
+    DO_NOT_CHECK_DND;
+
+    public static final String CHECK_DND_STRINGVAL = "S1";
```
koshalt: they set the service ids to s1 and s2? :expressionless:

PR #470, opened 2015-07-09
Fixes NMS-228

larubbio:
Allow State and District to be optional during csv import, but still reject with error.


koshalt:
lgtm. Unrelated to this checkin but are we moving to a model where the csv upload always succeeds with the errors being logged? Looks like we either fail on first error or log and keep moving in other places.

larubbio:
I don't think we have a consistent policy around that.  I think for csv
files from IMI we fail fast.  For mcts imports we log the error and
continue.  For others we fail fast.  I'm ok with that.

PR #471, opened 2015-07-09
Fixes NIP-57

larubbio:
If either MsgPlayTime field is missing return the play duration as 0.  These are optional fields.


koshalt:
lgtm

PR #474, opened 2015-07-09
137

kamknitt:
Needs to wait for http://review.motechproject.org/#/c/5398/ to be merged.

The changes tracking feature works only if the setter method is invoked on the instance that is attached to the PM. It may be the case only for updates. The object must be first retrieved from the data service and then modified via setters - all within a transaction.
Because of this limitation, changes tracking for updates made via MDS UI doesn't work at the moment.

### DIFF
--- a/flw/pom.xml
+++ b/flw/pom.xml
@@ -19,6 +19,7 @@
     </properties>
 
     <dependencies>
+
         <dependency>
             <groupId>org.motechproject</groupId>
             <artifactId>motech-platform-osgi-extender-fragment</artifactId>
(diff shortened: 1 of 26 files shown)


larubbio:
Fixes #377
Fixes #137

larubbio:
ok to test

larubbio:
lgtm.  Waiting for build to complete.

larubbio:
@kamknitt It looks like this commit is breaking some of the unit tests.  So those will need to be fixed before we can merge it.

kamknitt:
@larubbio I've fixed the unit tests. They should be ok now.

PR #477, opened 2015-07-09
Fixes #100

### LINKED ISSUE #100: NMS.GEN.LOC.004 - Method for end user to update location data
The NMS system shall provide means to correct details of location data through a web interface via CSV upload or MDS UI.

It should be possible to do changes in spelling or other details of location

larubbio:
Location data can be updated via csv upload.  Also contains a test for csv audit file name.


larubbio:
poop

This commit isn't actually this large.  I merged master into it (and rebased) which is why I think it is showing so much.  Only need to review the files in  region/src/main/java/org/motechproject/nms/region/csv/impl and the tests.  Or just take my word for it that it's good.

llavoie:
I'll take your word for it.

-----Original Message-----

larubbio:
Well then you'd be mistaken, because it's not just good ...

PR #478, opened 2015-07-09
Fixes NIP-34

larubbio:
This allows for the creation and association of States, Districts, Circles and Languages.  

I was able to create a subscriber, but the number of related tables made creation of all fields through the UI problematic.  I also think there may be an issue in MDS when creating a sub-class (I recieved an error trying to create MctsChild about no such field)

There isn't a requirement for creation of this data through the UI though so I don't feel like we need to fix all of those issues.


koshalt:
lgtm

PR #481, opened 2015-07-10
CR feedback from MCTS update work item

llavoie:
I haven't tested this or anything -- but I'm having the feeling I won't have a chance to finish it as it appears to be go time. :) If someone could take it over, that would be great.


frankhuster:
YAY!!!!

larubbio:
I'm sure it's just braxton hicks.  BACK TO WORK!!!!  :)

koshalt:
go time and you pushed out a review first? 
bdss

frankhuster:
Oh, and despite the minor contraction issue: lgtm!

PR #482, opened 2015-07-10
Implemented cases: 474, 475, 483, 484, 485, 486, 491, 499, 450

AnkitGaurNms:
Implemented cases: 474, 475, 483, 484, 485, 486, 491, 499, 450
Modified existing cases mapped to : 470, 471


koshalt:
looks mostly good. needs a merge from master though.

PR #484, opened 2015-07-10
Fixes NIP-65

larubbio:
Subscribers via IVR will not have an LMP date.  So we check if one exists before attempting
to update the start date.


frankhuster:
lgtm

PR #485, opened 2015-07-10
Fixes NIP-64

larubbio:
Given a subscribers district attempt to find their circle and assign it to them.


frankhuster:
Don't you **love** detatched fields? ;-)

frankhuster:
Any way we can write an IT for that?

frankhuster:
Other than that, lgtm.

larubbio:
The IT would load a user via csv and see if their circle was set.  I guess I could ......

PR #486, opened 2015-07-10
Here's your IT Frank!  

larubbio:
FYI this is failing for me locally, but I don't believe it

cc @frankhuster


frankhuster:
Don't remember what I asked for for a squared circle must be right, so lgtm.

PR #487, opened 2015-07-11
Fixes NIP-67

larubbio:
CallAnswerTime is optional in the CDR file so if it isn't provided don't return InvalidArgument

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/CsvHelper.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/CsvHelper.java
@@ -58,13 +58,23 @@ public final class CsvHelper {
     }
 
 
+    private static Long longOrNullFromString(String which, String s) {
+        if (which == null || which.isEmpty()) { return null; }
+
+        try {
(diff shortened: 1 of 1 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/CsvHelper.java
```diff
@@ -130,8 +130,13 @@ public static CallDetailRecordDto csvLineToCdrDto(String line) {
 
         cdr.setMsisdn(msisdnFromString(fields[FieldName.MSISDN.ordinal()]));
 
-        cdr.setCallAnswerTime(new DateTime(longFromString("CallAnswerTime",
-                fields[FieldName.CALL_ANSWER_TIME.ordinal()])));
+        try {
+            cdr.setCallAnswerTime(new DateTime(longFromString("CallAnswerTime",
```
frankhuster: This will throw IllegalArgumentException if the number is malformed and if it's null. How about making something like `longOrNullFromString()`?

frankhuster:
lgtm

PR #489, opened 2015-07-13
Added test-case

PriyaJain5:


larubbio:
This needs to have master merged into it.

PriyaJain5:
During Mcts_update upload if any exception occurs, Record is getting inserted in table nms_subscription_errors with beneficiaryId instead callingNum. So i need to add a method in SubscriptionErrorDataService.java to find the faulty record by beneficeryid. Please review java class file SubscriptionErrorDataService.java

PR #492, opened 2015-07-13
Fix for NIP-42-61-70

koshalt:
Changes Kilkari message retrieval to fix the bugs.


koshalt:
To clarify further,

NIP-42 claims that a missing message in the db for a scheduled message, should result in a standard error message for the get inbox api call, which we disagree with. This fix just implements a 'better' way to lookup a message for a given week. We will spit out error messages in logs for missing/bad deployments.

NIP-61 A potentially bad query result in a wrong subscription being picked for the day, followed by a fatal exception and non-generation of the OBD file. This fix will log the erroneous subscription and keep chugging. With the query fixed, this scenario should never be hit again though.

NIP-70 We never selected the second message for the pack on the OBD. This query fix will resolve that.

larubbio:
lgtm

PR #495, opened 2015-07-15
Fixes #430

### LINKED ISSUE #430: Import circles into nms
We need a way to upload/import circle data into the system.

larubbio:
Create a way to import circles

### DIFF
--- a/region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
+++ b/region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
@@ -1,0 +1,89 @@
+package org.motechproject.nms.region.csv.impl;
+
+import org.motechproject.nms.csv.exception.CsvImportDataException;
+import org.motechproject.nms.csv.utils.ConstraintViolationUtils;
+import org.motechproject.nms.csv.utils.CsvImporterBuilder;
+import org.motechproject.nms.csv.utils.CsvMapImporter;
+import org.motechproject.nms.csv.utils.GetString;
+import org.motechproject.nms.region.csv.CircleImportService;
+import org.motechproject.nms.region.domain.Circle;
+import org.motechproject.nms.region.domain.State;
+import org.motechproject.nms.region.repository.CircleDataService;
+import org.motechproject.nms.region.repository.StateDataService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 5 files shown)


region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    @Transactional
+    public void importData(Reader reader) throws IOException {
+        CsvMapImporter csvImporter = new CsvImporterBuilder()
+                .setProcessorMapping(getProcessorMapping())
+                .createAndOpen(reader);
+        Map<String, Object> record;
+        while (null != (record = csvImporter.read())) {
```
frankhuster: Don't we want null on the right? ;-)

region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+            try {
+                String circleName = (String) record.get(CIRCLE_NAME);
+                if (circleName == null) {
+                    throw new CsvImportDataException(String.format("CSV import error, circle missing on line %s",
```
frankhuster: Probably don't want to include "CSV import error, " since you're already saying it on line 77
Same comment below as well...

region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
larubbio: Honestly I just copied and pasted that code from either language or state
importer.  The best part is CPD didn't figure it out!

frankhuster: ha!

frankhuster:
lgtm; fix the nitpicks if you want, or ignore & merge.

region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
larubbio: That is added if ConstraintViolation is thrown.  Since I'm throwing CsvImportData it shouldn't be duplicated.

larubbio:
ignored and merged! :smiley:

region/src/main/java/org/motechproject/nms/region/csv/impl/CircleImportServiceImpl.java
frankhuster: Doh!

PR #503, opened 2015-07-16
Fixes NIP-72

larubbio:
When we create a subscription we now set the first and second message date.  When we load
subscriptions at target generation time we decide if we should load only the first message day or first and second message day.


koshalt:
lgtm

PR #511, opened 2015-07-20
Fixes for MA call details storage

koshalt:
Fixes NMS-232
Fixes NMS-233

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
@@ -27,6 +27,9 @@ import org.springframework.web.bind.annotation.ResponseStatus;
 public class CallDetailsController extends BaseController {
 
     public static final int MILLISECONDS_PER_SECOND = 1000;
+    private static final String QUESTION_TYPE = "question";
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
```diff
@@ -186,13 +193,31 @@ private String validateCallDetailsCommonElements(CallDetailRecordRequest callDet
...
         }
 
         return failureReasons.toString();
     }
 
+    private String validateCallDetailsMobileAcademyElements(CallDetailRecordRequest callDetailRecordRequest) {
+        StringBuilder failureReasons = new StringBuilder();
+
+
+
```
larubbio: Three blank lines?  Really?

really-seth-and-amy

larubbio:
lgtm other than the mind bogglingly dumb decision to use three blank lines in a row in a function.

api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
koshalt: Ah, I meant to write a comment there and forgot..but
escalated

PR #512, opened 2015-07-20
NIP-53

frankhuster:
Fixes NIP-53

Now inspecting Call Summary Report files in order to retry failed calls not reported in Call Detail Report files


frankhuster:
Looks like it's not completely fixed. Two things need to happen:

1 - I need to make things work
2- I need to add an IT to check I did fix the fix!

I'll post back with a new PR.

frankhuster:
@koshalt, it's all yours!

koshalt:
cheering_minions

imi/src/main/java/org/motechproject/nms/imi/domain/CallSummaryRecord.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public Integer getAttempts() {
+        return attempts;
+    }
+
+    public void setAttempts(Integer attempts) {
+        this.attempts = attempts;
+    }
+
+    public CallSummaryRecordDto toDto() {
```
koshalt: Maybe a nitpick, but I think this would make more sense in the Dto than here. The single truth of the domain object can be translated into (and by) multiple dto objects, rather than this domain object knowing about Dto(s)

imi/src/main/java/org/motechproject/nms/imi/service/CdrFileService.java
```diff
@@ -22,39 +23,39 @@
      * Verifies the checksum & record count provided in fileInfo match the checksum & record count of file
      * also verifies all csv rows are valid.
      *
-     * @param file          file to process
      * @param fileInfo      file information provided about the file (ie: expected checksum & recordCount)
```
koshalt: Do the comments need to be updated with the new param too?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -278,25 +281,31 @@ private CallSummaryRecord aggregateSummaryRecord(CallSummaryRecordDto csr) {
...
+                break;
+
+            case REJECTED:
+                deactivateSubscription(subscription, callRetry);
+                break;
 
-        if (csr.getFinalStatus() == FinalCallStatus.REJECTED) {
-            deactivateSubscription(subscription, callRetry);
+            default:
+                throw new IllegalStateException(String.format("Invalid FinalCallStatus: %s", csrDto.getFinalStatus()));
```
koshalt: Its probably better to just log/alert and fail here since an exception on the handler puts the event back on the queue to be picked up by someone else.

koshalt:
Should be good to go after the comments are addressed.

PR #522, opened 2015-07-23
Added test cases for flw.

PriyaJain5:


testing/src/test/java/org/motechproject/nms/testing/it/api/UserControllerBundleIT.java
```diff
@@ -5171,4 +5171,108 @@ public void verifyFT534() throws IOException, InterruptedException {
...
+
+    HttpPost httpPost = createHttpPost("mobilekunji", new UserLanguageRequest(1200000001L, 123456789012346L,
+            rh.hindiLanguage().getCode()));
+
+    response = SimpleHttpClient.httpRequestAndResponse(httpPost, ADMIN_USERNAME, ADMIN_PASSWORD);
+
+    assertEquals(HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+
+    flw = frontLineWorkerService.getByContactNumber(1200000001l);
+    assertEquals(FrontLineWorkerStatus.ACTIVE, flw.getStatus());
```
larubbio: As discussed in https://applab.atlassian.net/browse/NMS-246 we don't set the user to active until we get the call details back from IMI.  So this test isn't valid.  If you update it to simulate the call details and then check the user it will pass.

PR #525, opened 2015-07-23
For NIP-75

larubbio:
I'm not sure if this will completely fix it.  I'm just checking in so the CI will run all the unit tests while I commute to the office.

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/UserController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/UserController.java
@@ -30,9 +30,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseBody;
 
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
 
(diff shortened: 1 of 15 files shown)


larubbio:
ok to test

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -137,7 +135,7 @@ public UserResponse getUserDetails(@PathVariable String serviceName,
 
         // If the user does not have a language location code we want to return the allowed language location
         // codes for the provided circle, or all if no circle was provided
-        List<Language> languages = new ArrayList<>();
+        Set<Language> languages = new HashSet<>();
         if (user.getLanguageLocationCode() == null && circleObj != null) {
```
koshalt: Unrelated to this, but intellij warns of a NPE here since there is a possibility that the user object was never set.

larubbio: Yeah.  I was thinking of adding an assert above.

First we verify we get one of the three possible services, if not we bail.
Then we call either getFLWUser or getKilkariUser, both of which will either
throw or return a user.  So it isn't actually possible.

larubbio:
ok to test

koshalt:
lgtm

PR #526, opened 2015-07-23
Fix race condition between db & activemq

### LINKED ISSUE #518: Race condition in MA controller and message handler
For saving bookmarks, a db instance is created and an event trigger in case of course completion. However, the db instance happens inside of a transaction while the fire event (and handling) happens right away. This causes the handler to not find the db instance for the completion record and fail.

Solutions:
1. Roll the JMS transaction (event fire) into the spring db transaction
2. Setup scheduler to fire ~5-10secs after the call to guarantee that the processing is done.

koshalt:
Fixes #518


larubbio:
lgtm pending build agreeing with me.

PR #527, opened 2015-07-23
Added check for flw state transition

koshalt:
Fixes the symptoms reported in NMS-246, although the expected workflow in the bug is incorrect.

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
@@ -51,7 +51,7 @@ public class CallDetailsController extends BaseController {
      * /api/mobilekunji/callDetails
      *
      */
-    @RequestMapping(value = "/{serviceName}/callDetails",
+    @RequestMapping(value = "/{serviceName}/callDetails", // NO CHECKSTYLE Cyclomatic Complexity
             method = RequestMethod.POST,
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/motechproject/nms/api/web/CallDetailsController.java
```diff
@@ -106,13 +106,23 @@ public void saveCallDetails(@PathVariable String serviceName,
 
         createCallDetailRecord(flw, callDetailRecordRequest, service);
 
-        // if this is the FLW's first time calling the service, set her status to ACTIVE
-        if (flw.getStatus() == FrontLineWorkerStatus.INACTIVE) {
+        // if this is the FLW's first time calling the service, set her status to ACTIVE based on NMS.GEN.FLW.003
+        if (flw.getStatus() == FrontLineWorkerStatus.INACTIVE &&
+                validateFlwNameAndNumber(flw) &&
+                validateFlwLocation(flw)) {
```
larubbio: According to the spec we also need valid designation.  GET ON IT!

koshalt: oo

larubbio:
lgtm.  Waiting for build

larubbio:
ok to test

larubbio:
ok to test

larubbio:
ok to test

PR #529, opened 2015-07-24
Security settings

srivastavaAbhi:
Update config file

With initial pull request #507   only configuration present in FLW module is coming in dynamic URL setting.
Now i have updated "origin" in each files from "all" to "SYSTEM_MODULE_<modulename>" and luckily i get all the entries in dynamic URL list.


larubbio:
ok to test

PR #543, opened 2015-07-28
1 off mistake when dealing with more than 1000 fresh calls...

frankhuster:


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -478,6 +480,8 @@ public Subscription create(Subscription subscription) {
         return subscriptionDataService.executeQuery(queryExecution);
     }
 
+    //todo: IT to make sure chunking works!
```
koshalt: chunking...lol

PR #547, opened 2015-07-29
For NIP-82

larubbio:
There was confusion since we have callStatus and finalCallStatus in the database.  I don't remember why we have finalCallStatus, but rather then risk breaking something else I'm just preventing MDS from generating a column for callStatus since we don't have any value we will store there.


koshalt:
test this please

larubbio:
ok te test

larubbio:
ok to test

larubbio:
ok to test

koshalt:
@larubbio are you ok with this change?

larubbio:
lgtm

PR #550, opened 2015-08-03
Modify FT 432

PriyaJain5:
JIRA: https://applab.atlassian.net/browse/NMS-256


larubbio:
This will be a spec change so that FT isn't valid.

PR #555, opened 2015-08-03
Fixes NMS-254

larubbio:
Verify address is not null before obscuring it

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/sms/DeliveryInfo.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/sms/DeliveryInfo.java
@@ -36,13 +36,13 @@ public class DeliveryInfo {
     }
 
     private static String obscure(String a) {
-        return "tel: *******" + a.substring(a.length() - MASK_LENGTH);
+        return a == null ? "null" : "tel: *******" + a.substring(a.length() - MASK_LENGTH);
     }
 
     @Override
     public String toString() {
         return "DeliveryInfo{" +
-                "address='" + obscure(address) + '\'' +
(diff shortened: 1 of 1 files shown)


larubbio:
ok to test

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/sms/DeliveryInfo.java
```diff
@@ -36,13 +36,13 @@ public void setDeliveryStatus(DeliveryStatus deliveryStatus) {
...
     private static String obscure(String a) {
-        return "tel: *******" + a.substring(a.length() - MASK_LENGTH);
+        return a == null ? "null" : "tel: *******" + a.substring(a.length() - MASK_LENGTH);
     }
 
     @Override
     public String toString() {
         return "DeliveryInfo{" +
-                "address='" + obscure(address) + '\'' +
+                "address='" +  obscure(address) + '\'' +
```
frankhuster: Double space?!? OMG YOU'RE CRAZY!!!

frankhuster:
lgtm

frankhuster:
grrr this `testTargetFileGeneration` failure is annoying

frankhuster:
ok to test

api/src/main/java/org/motechproject/nms/api/web/contract/mobileAcademy/sms/DeliveryInfo.java
larubbio: I originally had the null check there, but then figured it would be better
in the obscure method.  I guess an extra ' ' didn't want to be deleted.
Poor little guy, can we keep him?

PR #556, opened 2015-08-05
Fixes NIP-106

larubbio:
I think jackson is trying to serialiaze a getter that we don't need it to.  I annotated it with @JsonIgnore


frankhuster:
What is that fixing?

PR #557, opened 2015-08-05
Fixes NIP-105

larubbio:
We now log the response as well as the request on API calls.

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
@@ -88,7 +88,7 @@ public class KilkariController extends BaseController {
     public InboxResponse getInboxDetails(@RequestParam(required = false) Long callingNumber,
                                          @RequestParam(required = false) Long callId) {
 
-        log("/kilkari/inbox", String.format("callingNumber=%s, callId=%s",
+        log("REQUEST: /kilkari/inbox", String.format("callingNumber=%s, callId=%s",
                 LogHelper.obscure(callingNumber), callId));
 
         StringBuilder failureReasons = validate(callingNumber, callId);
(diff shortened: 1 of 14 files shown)


api/src/main/java/org/motechproject/nms/api/web/contract/FlwUserResponse.java
```diff
@@ -68,6 +68,17 @@ public void setMaxAllowedEndOfUsagePrompt(int maxAllowedEndOfUsagePrompt) {
...
 
     @Override
+    public String toString() {
+        return "FlwUserResponse{" +
+                "currentUsageInPulses=" + currentUsageInPulses +
+                ", endOfUsagePromptCounter=" + endOfUsagePromptCounter +
+                ", welcomePromptFlag=" + welcomePromptFlag +
+                ", maxAllowedUsageInPulses=" + maxAllowedUsageInPulses +
+                ", maxAllowedEndOfUsagePrompt=" + maxAllowedEndOfUsagePrompt +
+                "} " + super.toString();
```
frankhuster: I never saw a super.toString inside a toString before, cool. and a bit weird.

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
```diff
@@ -88,7 +88,7 @@
     public InboxResponse getInboxDetails(@RequestParam(required = false) Long callingNumber,
                                          @RequestParam(required = false) Long callId) {
 
-        log("/kilkari/inbox", String.format("callingNumber=%s, callId=%s",
+        log("REQUEST: /kilkari/inbox", String.format("callingNumber=%s, callId=%s",
```
frankhuster: If we're marking posts with (POST), should we also mark gets with (GET)?

frankhuster:
lgtm, one suggestion you're welcome to follow or not, then merge yourself.

api/src/main/java/org/motechproject/nms/api/web/contract/FlwUserResponse.java
larubbio: intellij has that as one of the options when you tell it to insert
toString()

api/src/main/java/org/motechproject/nms/api/web/KilkariController.java
larubbio: One method had that in it's description I removed it and decided to leave
GET of as the default for no good reason.

PR #562, opened 2015-08-16
Debug output and slightly different cache loading

frankhuster:


larubbio:
The commit message also mentions cache loading changes but I don't see that.  Should that be included?

frankhuster:
Setting the globals to null will force a reload.

larubbio:
lgtm

PR #563, opened 2015-08-16
Better CDRFileProcessedNotification

frankhuster:
Manjot wondered if we sent notification in case of errors.
I checked, and while it looks like we did for most, we didn't for all.
So I changed the logic.

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
@@ -619,18 +619,7 @@ public class CdrFileServiceImpl implements CdrFileService {
     }
 
 
-    // Phase 2: Aggregate detail records into summary records
-    //          Validate all CallSummaryRecord fields exist in the database
-    //          Validate CallSummaryFile
-    //          Stop processing if errors occurred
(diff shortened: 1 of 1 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -718,18 +727,20 @@ public void verifyDetailFileChecksumAndCount(CdrFileNotificationRequest request)
...
+        File cdrFile = new File(localCdrDir(), request.getCdrDetail().getCdrFile());
         errors = sendAggregatedRecords(cdrFile);
         if (errors.size() > 0) {
             reportAuditAndPost(request.getCdrDetail().getCdrFile(), errors);
+            return errors;
         }
 
 
         // Phase 4: distribute the processing of CSRs to all nodes, make backup copy of all received CSR
+        File csrFile = new File(localCdrDir(), request.getCdrSummary().getCdrFile());
```
koshalt: is this supposed to be csr? copy-paste error?

koshalt: nvm, cleared up on the :iphone:

PR #566, opened 2015-08-18
Build sms content for completion

### LINKED ISSUE #284: Host and serve localized sms content for course completion
Based on recent thread with BBC - 

Looks like we will need to host the sms content for all the different language codes (transliterated to english) and send the course completion sms notification with this content.

TODO -
1. ingest and store content (based on llc)
2. Based on the users llc at completion, choose the appropriate content to send and call IMI

koshalt:
Fixes #284 
Not the most elegant solution with having the content in a properties file since they would ideally live in IMI's cms system. This is a start that can be optimized further though.

### DIFF
--- a/mobile-academy/src/main/resources/sms-content.properties
+++ b/mobile-academy/src/main/resources/sms-content.properties
@@ -1,0 +1,5 @@
+# localized content for completion sms (by language code)
+# TODO: update the language code suffix wherever necessary, assuming that 50 is hindi (national default)
+# TO add more languages, add as sms.content.61 or sms.content.72 where 61 and 72 would be the language codes
+# used in the deployment. (tamizh & punjabi for example)
+sms.content.50 = "Badhai ho didi, apne Mobile Academy course pura kar liya hai. Certificate paane ke liye is SMS ko apne subcenter meeting mein dikhayen. No "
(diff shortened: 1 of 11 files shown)


mobile-academy/src/main/resources/sms-content.properties
```diff
@@ -0,0 +1,5 @@
+# localized content for completion sms (by language code)
+# TODO: update the language code suffix wherever necessary, assuming that 50 is hindi (national default)
+# TO add more languages, add as sms.content.61 or sms.content.72 where 61 and 72 would be the language codes
+# used in the deployment. (tamizh & punjabi for example)
+sms.content.50 = "Badhai ho didi, apne Mobile Academy course pura kar liya hai. Certificate paane ke liye is SMS ko apne subcenter meeting mein dikhayen. No "
```
larubbio: Maiṁ hinda bāta nahīṁ hai, lēkina mujhē lagatā hai ki yadi maiṁ anta mēṁ atirikta antarikṣa kē bārē mēṁ śikāyata karēṅgē.

larubbio:
मुझे अच्छा लग रहा है

koshalt:
நன்றி

mobile-academy/src/main/resources/sms-content.properties
koshalt: What is this? Bihari? Sounds like hindi but I have no idea..

larubbio: It's whatever google translate came up with.  It's supposed to say something along the lines of "I don't speak hindi, but if I did I'd be concerned with the extra space at the end."

koshalt: Ah ok. That space was specified by BBC as the format for the location number (No xxxxxxx) correlation.

PR #567, opened 2015-08-19
Fixes NMS-236

larubbio:
When a user is marked as INVALID we now null out their contact number.  The non-null
constraint in the database has been removed but a class level validator has been added that
enforces a contact number when the status is not INVALID and that contact number is null when
status is INVALID.

Since the INVALID FLW has no contact number there will be no conflict if a user with that number calls MA/MK.

All other information is retained in the record.

### DIFF
--- a/flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
@@ -41,7 +41,7 @@ public class FrontLineWorker extends MdsEntity implements FullLocation {
     @Field
     private String mctsFlwId;
 
-    @Field(required = true)
+    @Field()
     @Unique
     @Min(value = 1000000000L, message = "contactNumber must be 10 digits")
     @Max(value = 9999999999L, message = "contactNumber must be 10 digits")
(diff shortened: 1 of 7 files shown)


flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
```diff
@@ -41,7 +41,7 @@
     @Field
     private String mctsFlwId;
 
-    @Field(required = true)
+    @Field()
```
koshalt: nitpick is as nitpick does, () not needed?

testing/src/test/java/org/motechproject/nms/testing/it/flw/FrontLineWorkerServiceBundleIT.java
```diff
@@ -266,7 +265,7 @@ public void testDeleteRecentInvalidFrontLineWorker() {
         flw.setStatus(FrontLineWorkerStatus.INVALID);
         frontLineWorkerService.update(flw);
 
-        flw = frontLineWorkerService.getByContactNumber(2111111111L);
+        flw = frontLineWorkerService.getById(flw.getId());
         assertEquals(FrontLineWorkerStatus.INVALID, flw.getStatus());
```
koshalt: should we also add an assert here for contact number being null after the update?

koshalt:
sembra buono per me

PR #568, opened 2015-08-19
Potential fix for NIP-140

larubbio:
This constructs an SSLConnectionFactory with TLS v1, v1.1 & v1.2 enabled.


larubbio:
#fingerscrossed

manjotnms:
Now we are facing problem with peer domain name verification, As now we are using CA signed certificate.

2015-09-01 09:21:01,379 WARN  [org.motechproject.nms.imi.service.impl.ExponentialRetrySender] Unable to send httpPost POST https://*.nationalmhealth.in/obdmanager/notifytargetfile HTTP/1.1: Host name '*.nationalmhealth.in' does not match the certificate subject provided by the peer (CN=*.nationalmhealth.in, O=IMI MOBILE PRIVATE LIMITED, L=Hyderabad, C=IN)

Problem seems to with hostname verification refer to the below link

Refer to the link
http://stackoverflow.com/questions/29124161/how-come-apache-httpclient-4-4-rejects-www-googleapis-com-as-a-valid-host-name  

Should we use NoopHostnameVerifier? or any other suggestions.

PR #569, opened 2015-08-21
Perf

frankhuster:
Probably as good a time as ever to push this one.
Maybe we can have the changes to python+data in a new branch and the java+testing changes in yet another?


kilkari/src/main/java/org/motechproject/nms/kilkari/util/Timer.java
```diff
@@ -0,0 +1,64 @@
+package org.motechproject.nms.kilkari.util;
```
koshalt: should we put this in a more common package for use with other modules for perf testing?

performance/data/data.py
```diff
@@ -0,0 +1,93 @@
+import argparse
```
koshalt: :snake:

performance/models.py
```diff
@@ -0,0 +1,746 @@
...
+    endDate = Column(DateTime)
+    firstMessageDayOfWeek = Column(String(255, u'utf8_bin'))
+    modificationDate = Column(DateTime, nullable=False)
+    modifiedBy = Column(String(255, u'utf8_bin'), nullable=False)
+    needsWelcomeMessageViaObd = Column(BIT(1), nullable=False)
+    origin = Column(String(255, u'utf8_bin'), nullable=False)
+    owner = Column(String(255, u'utf8_bin'))
+    secondMessageDayOfWeek = Column(String(255, u'utf8_bin'))
+    startDate = Column(DateTime, index=True)
+    status = Column(String(255, u'utf8_bin'), nullable=False, index=True)
```
koshalt: are these always 255 by default or is this model different from what gets generated by MDS? This is more in the context of me wondering if the model will affect the perf testing in any way. I trust you though

larubbio: I don't think it should make a difference.  I figure this is just for inserting data.  This model was probably generated by looking at the database.

testing/src/main/java/org/motechproject/nms/testing/service/impl/TestingServiceImpl.java
```diff
@@ -174,100 +212,481 @@ public TestingServiceImpl() {
     }
 
 
+    private void truncateTable(final String table) {
+        SqlQueryExecution sqe = new SqlQueryExecution() {
+
+            @Override
+            public String getSqlQuery() {
+                return String.format("DELETE FROM %s WHERE 1=1", table);
```
koshalt: nice. but what happens when 1!=1 :laughing:

frankhuster: It's a trick to avoid getting an "you can't do a global delete on a table without a where clause" error.

kilkari/src/main/java/org/motechproject/nms/kilkari/util/Timer.java
frankhuster: Yep, I was wondering where.

testing/src/test/java/org/motechproject/nms/testing/it/imi/TargetFileServiceBundleIT.java
```diff
@@ -274,7 +274,7 @@ public void createLargeFile() {
             sh.mksub(SubscriptionOrigin.MCTS_IMPORT, DateTime.now());
         }
 
-        for (int i=0 ; i<1000 ; i++) {
+        for (int i=0 ; i<0 ; i++) {
```
koshalt: is this a typo?

kilkari/src/main/java/org/motechproject/nms/kilkari/util/Timer.java
koshalt: maybe in a separate module called performance? (assuming that you don't run into circular dependencies)

testing/src/test/java/org/motechproject/nms/testing/it/imi/TargetFileServiceBundleIT.java
frankhuster: nope, this is not run anyways

PR #575, opened 2015-08-31
Test for fetching all language codes

motech-gerrit:


motech-gerrit:

PR #584, opened 2015-09-09
Faster get user

frankhuster:
Uses a JDO query which avoids a SELECT WHERE EXISTS ending up in a table scan.


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -53,6 +55,8 @@
     private static final String WEEKS_TO_KEEP_CLOSED_SUBSCRIPTIONS = "kilkari.weeks_to_keep_closed_subscriptions";
 
     private static final String SUBSCRIPTION_PURGE_EVENT_SUBJECT = "nms.kilkari.purge_closed_subscriptions";
+    public static final String SELECT_SUBSCRIBERS_BY_NUMBER = "select * from nms_subscribers where callingNumber = ?";
```
koshalt: do you think that we can pull this into a common static class with queries defined for better manageability?

frankhuster: Sure, or keep the queries inline in the code - I did that here to work around CPD (grr). Both the Subscriber service and Subscription service have the same getUser method. Your choice.

koshalt: meh.. I guess we can do this in the future. more of a nitpick than anything else.

koshalt:
lgtm

PR #587, opened 2015-09-12
Bug fix for sms content generation

koshalt:
Fixes NIP-152

### DIFF
--- a/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
+++ b/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
@@ -39,7 +39,7 @@ public class CourseNotificationServiceImpl implements CourseNotificationService 
     private static final String DELIVERY_STATUS = "deliveryStatus";
     private static final String ADDRESS = "address";
     private static final String SMS_CONTENT_PREFIX = "sms.content.";
-    private static final String SMS_DEFAULT_LANGUAGE_CODE = "default";
+    private static final String SMS_DEFAULT_LANGUAGE_PROPERTY = "default";
     private static final Logger LOGGER = LoggerFactory.getLogger(CourseNotificationServiceImpl.class);
 
     /**
@@ -187,7 +187,7 @@ public class CourseNotificationServiceImpl implements CourseNotificationService 
     private String buildSmsContent(Long callingNumber) {
 
         FrontLineWorker flw = frontLineWorkerService.getByContactNumber(callingNumber);
-        String locationId = "XX"; // unknown location id
+        String locationCode = "XX"; // unknown location id
         String smsLanguage = null;
 
         if (flw == null) {
@@ -196,7 +196,7 @@ public class CourseNotificationServiceImpl implements CourseNotificationService 
 
         // Build location code
         if (flw.getState() != null && flw.getDistrict() != null) {
-            locationId = flw.getState().getCode().toString() + flw.getDistrict().getCode();
+            locationCode = flw.getState().getCode().toString() + flw.getDistrict().getCode();
         }
 
         // set sms content language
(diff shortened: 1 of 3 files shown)


mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
```diff
@@ -188,26 +187,36 @@ public void updateSmsStatus(MotechEvent event) {
...
+
         if (flw == null) {
             throw new IllegalStateException("Unable to find flw for calling number: " + callingNumber);
         }
-        Language flwLanguage = flw.getLanguage() != null ? flw.getLanguage() : flw.getDistrict().getLanguage();
-        String smsLanguage;
-        if (flwLanguage != null && flwLanguage.getCode() != null) {
-            smsLanguage = flwLanguage.getCode();
+
+        // Build location code
```
frankhuster: location id or location code?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
```diff
@@ -40,7 +39,7 @@
     private static final String DELIVERY_STATUS = "deliveryStatus";
     private static final String ADDRESS = "address";
     private static final String SMS_CONTENT_PREFIX = "sms.content.";
-    private static final String SMS_DEFAULT_LANGUAGE_CODE = "50";
+    private static final String SMS_DEFAULT_LANGUAGE_CODE = "default";
```
frankhuster: that's the name of the property, not the value, right?
wanna rename it to show that?

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
```diff
@@ -409,6 +409,31 @@ public void testNotification() {
         // TODO: cannot check the notification status yet since we don't have a real IMI url to hit
     }
 
+    @Test
+    public void testNotificationNoLocation() {
+        // Setup language/location and flw for notification
```
frankhuster: where do you setup the languag/location?

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
```diff
@@ -409,6 +409,31 @@ public void testNotification() {
...
+            frontLineWorkerService.update(flw);
+            frontLineWorkerService.delete(flw);
+        }
+
+        flw = new FrontLineWorker("Test Worker", 2111113333L);
+        frontLineWorkerService.add(flw);
+        flw = frontLineWorkerService.getByContactNumber(2111113333L);
+        assertNotNull(flw);
+
+        long callingNumber = 2111113333L;
```
frankhuster: If you're using 2111113333L everywhere and you're using a variable, why not use that same variable in the entire test instead of using the literal? Or, am I missing something?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
koshalt: it is location identifier string that needs to be built.. doesn't matter what we call it here. Open to whatever sounds good.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
koshalt: not sure what this means, but the property is call sms.content.default. What do you recommend we call it?

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
koshalt: copy-paste issue. fixed

testing/src/test/java/org/motechproject/nms/testing/it/ma/MobileAcademyServiceBundleIT.java
koshalt: fixed.

koshalt:
Everything's been fixed. @frankhuster come 'git' it

PR #590, opened 2015-09-15
Implement caching using Spring's @Cacheable

frankhuster:


frankhuster:
ok to test

tracking/src/main/java/org/motechproject/nms/tracking/service/CacheService.java
```diff
@@ -0,0 +1,5 @@
+package org.motechproject.nms.tracking.service;
+
+public interface CacheService {
```
larubbio: Is there a reason you created the cache service bean?  My concern is if we want to cache the results of another method we should just be able to add the @Cacheable annotation and not need to add a new method to this bean

frankhuster: @Cacheable won't work on internal methods.

http://stackoverflow.com/questions/12115996/spring-cache-cacheable-method-ignored-when-called-from-within-the-same-class

On Wed, Sep 16, 2015 at 11:07 AM, Rob LaRubbio <notifications@github.com>
wrote:

> In
> tracking/src/main/java/org/motechproject/nms/tracking/service/CacheService.java
> <https://github.com/motech-implementations/mim/pull/590#discussion_r39664464>
> :
>
> > @@ -0,0 +1,5 @@
> > +package org.motechproject.nms.tracking.service;
> > +
> > +public interface CacheService {
>
> Is there a reason you created the cache service bean? My concern is if we
> want to cache the results of another method we should just be able to add
> the @Cacheable annotation and not need to add a new method to this bean
>
> —
> <https://github.com/motech-implementations/mim/pull/590/files#r39664464>.
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

larubbio: ok, but this still feels like the wrong solution.  I think the correct fix would be to add cacheing to the platform on the entityService.getEntityByClassName method.  I don't see us doing that.

Is there an existing service in the tracking module that we could add that method to that would make sense?

frankhuster: I imagine we'll find one more place in the tracking module to cache, and
we'd then be adding one extra method to the CacheService class.
And I also imagine we'll do the same (create a CacheService class) in
kilkari and possibly other modules.

In the particular case of getEntityByClassName, you're right, it may very
well be added to the platform instead of being cached in this module.
But I want to proceed with perf.

Would it make sense to do a one pass after we're done optimizing and move
to platform what needs to be moved?

On Wed, Sep 16, 2015 at 11:58 AM, Rob LaRubbio <notifications@github.com>
wrote:

> In
> tracking/src/main/java/org/motechproject/nms/tracking/service/CacheService.java
> <https://github.com/motech-implementations/mim/pull/590#discussion_r39671014>
> :
>
> > @@ -0,0 +1,5 @@
> > +package org.motechproject.nms.tracking.service;
> > +
> > +public interface CacheService {
>
> ok, but this still feels like the wrong solution. I think the correct fix
> would be to add cacheing to the platform on the
> entityService.getEntityByClassName method. I don't see us doing that.
>
> Is there an existing service in the tracking module that we could add that
> method to that would make sense?
>
> —
> <https://github.com/motech-implementations/mim/pull/590/files#r39671014>.
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

larubbio: It sounds like you could get access to the proxy and call through that:

```java
((TrackChangesService)AopContext.currentProxy())).isEntityInstance(arg)
```

Although that is frowned against since it couples you to the spring framework.  

I don't think we should be creating catchall CacheServices and moving methods that we want cached into them.  If we have to do it here then I'll grumpily accept, but we shouldn't do it in kilkari and all the other places we want to cache the results of a method.

larubbio: Or if we keep the new service, instead of calling it CacheService I think we should rename it the EntityServiceWrapper and not expose it.  So instead of placing it in the service package we should either see if we've done something like this before and copy that, or create a new package internal/service so it isn't exposed via osgi.

frankhuster: I'll look into it this morning. And also will check if we can cache
repository lookups directly.

On Wed, Sep 16, 2015 at 4:53 PM, Rob LaRubbio <notifications@github.com>
wrote:

> In
> tracking/src/main/java/org/motechproject/nms/tracking/service/CacheService.java
> <https://github.com/motech-implementations/mim/pull/590#discussion_r39699290>
> :
>
> > @@ -0,0 +1,5 @@
> > +package org.motechproject.nms.tracking.service;
> > +
> > +public interface CacheService {
>
> Or if we keep the new service, instead of calling it CacheService I think
> we should rename it the EntityServiceWrapper and not expose it. So instead
> of placing it in the service package we should either see if we've done
> something like this before and copy that, or create a new package
> internal/service so it isn't exposed via osgi.
>
> —
> <https://github.com/motech-implementations/mim/pull/590/files#r39699290>.
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

frankhuster:
ok to test

frankhuster:
ok to test

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -140,6 +143,13 @@ public State getState(FrontLineWorker frontLineWorker) {
             if (language != null) {
                 Set<State> states = languageService.getAllStatesForLanguage(language);
 
+                //todo remove
```
larubbio: Should this be removed?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerUpdateImportServiceImpl.java
```diff
@@ -43,7 +43,7 @@
 
     private FrontLineWorkerService frontLineWorkerService;
     private LanguageService languageService;
-    private StateDataService stateDataService;
+    private StateService stateDataService;
```
larubbio: Rename this variable be to stateService.

flw/pom.xml
```diff
@@ -121,6 +121,7 @@
                             org.springframework.transaction,
                             org.apache.commons.fileupload,
                             org.springframework.web.multipart.commons,
+                            org.motechproject.nms.region.repository,
```
larubbio: Why was the dependency on region.repository added.  It looks like it already existed and you just added a new dependency on service.  Can this be removed?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -423,41 +422,29 @@ public void deactivateSubscription(Subscription subscription, DeactivationReason
     }
 
     @Override
+    @Cacheable("pack-type")
```
larubbio: Is it possible to combine the three subscription pack caches into one?  It probably isn't a big deal memory wise, but if someone were to copy this pattern for a entity with more objects then that could cause memory pressure.  

Either that or if in the cache config we set these to be small caches (like 2 objects each then we should be fine)

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
```diff
@@ -537,8 +524,9 @@ public Subscription create(Subscription subscription) {
     }
 
     public List<Subscription> findPendingSubscriptionsFromDate(DateTime startDate, int page, int pageSize) {
-        return subscriptionDataService.findByStatusAndStartDate(SubscriptionStatus.PENDING_ACTIVATION, startDate,
-                new QueryParams(page, pageSize));
+        return subscriptionDataService.findByStatusAndStartDate(SubscriptionStatus.PENDING_ACTIVATION, startDate, new QueryParams(page, pageSize));
     }
 
+    @CacheEvict(value = {"pack-type", "pack-name", "pack-all" }, allEntries = true)
```
larubbio: We still will need to expire the cache.  This will evict the entries on the server were the new pack was created, but the other servers will still have stale cached entries.

kilkari/src/main/resources/META-INF/motech/applicationContext.xml
```diff
@@ -31,4 +40,17 @@
             </list>
         </property>
     </bean>
+
+    <cache:annotation-driven />
+
+    <bean id="cacheManager" class="org.springframework.cache.support.SimpleCacheManager">
```
larubbio: Should we stick with SimpleCacheManager or move to ehcache now?

region/src/main/java/org/motechproject/nms/region/csv/impl/StateImportServiceImpl.java
```diff
@@ -30,6 +31,7 @@ public StateImportServiceImpl(StateDataService stateDataService) {
     }
 
     @Override
+    @CacheEvict(value = {"state-code", "state-name", "state-all" }, allEntries = true)
```
larubbio: It still feels wrong to have three caches for each object.  IT would be nice if we could build the key with a prefix so there would be only one to deal with. Not a big deal though

region/src/main/java/org/motechproject/nms/region/csv/impl/TalukaImportServiceImpl.java
```diff
@@ -35,10 +35,10 @@
 
     private TalukaService talukaService;
     private DistrictService districtService;
-    private StateDataService stateDataService;
+    private StateService stateDataService;
```
larubbio: Rename this to stateService

region/src/main/java/org/motechproject/nms/region/service/impl/DistrictServiceImpl.java
```diff
@@ -21,14 +25,19 @@
...
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(DistrictServiceImpl.class);
+
     @Autowired
     public DistrictServiceImpl(DistrictDataService districtDataService) {
         this.districtDataService = districtDataService;
     }
 
     @Override
+    @Cacheable("district-language")
```
larubbio: I think we can combine these per-method caches into a single cache using a custom key:

http://stackoverflow.com/a/14077831/379354

region/src/main/java/org/motechproject/nms/region/service/impl/DistrictServiceImpl.java
```diff
@@ -107,13 +124,23 @@ public District execute(Query query) {
     }
 
     @Override
-    public void create(District district) {
-        districtDataService.create(district);
+    @CacheEvict(value = {"district-state-name", "district-state-code", "district-language" }, allEntries = true)
+    public District create(District district) {
```
larubbio: Is there a reason eviction is done as an annotation on create/update methods as opposed to a lifecycle event that was used elsewhere?

tracking/pom.xml
```diff
@@ -75,6 +75,11 @@
                             org.motechproject.nms.tracking.utils;version=${project.version}
                         </Export-Package>
                         <Import-Package>
+                            org.aopalliance.aop,
```
larubbio: are the aop imports needed?  They weren't added above.

tracking/src/main/java/org/motechproject/nms/tracking/cache/CacheHelper.java
```diff
@@ -0,0 +1,5 @@
+package org.motechproject.nms.tracking.cache;
+
+public interface CacheHelper {
```
larubbio: This class should be called EntityServiceWrapper since that is what it is.  Spring provides the cache help for us.

larubbio:
I think if:
  - The per-method caches can be collapsed into per-entity caches using custom key generation
  - CacheHelper renamed to EntityServiceWrapper 

Then it is good to go.  There was some other small nitpicks (rename stateDataService) and we should discuss the move to ehcache.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
frankhuster: Yeah I just thought about that in the car on the drive back.
Don't we have a way for all nodes to subscribe to a special broadcast-like motech event?

larubbio: Yes, in serverEventRelay instead of calling sendEventMessage
call broadcastEventMessage although I'm not sure if your subscribe to that
the same way you do to a topic message.

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
frankhuster: :-)

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerUpdateImportServiceImpl.java
frankhuster: done

kilkari/src/main/resources/META-INF/motech/applicationContext.xml
frankhuster: I guess the good thing is doesn't really matter just yet as long as we get the annotations right.
#notansweringthequestion

flw/pom.xml
frankhuster: Probably trying to play with something. Removed, thanks.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
frankhuster: As of now, I'm only caching domain entities. I imagine we'd want to address that if we were trying to cache non domain entities, right?

tracking/pom.xml
frankhuster: I get the following error without them:
```
2015-09-23 09:49:46,629 EclipseGeminiBlueprintExtenderThread-31 ERROR [org.eclipse.gemini.blueprint.extender.internal.activator.ContextLoaderListener] Application context refresh failed (MotechOsgiConfigurableApplicationContext(bundle=org.motechproject.nms.tracking, config=osgibundle:/META-INF/spring/*.xml))
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityServiceWrapper' defined in URL [bundle://90.0:1/org/motechproject/nms/tracking/service/impl/EntityServiceWrapperImpl.class]: Initialization of bean failed; nested exception is java.lang.IllegalArgumentException: interface org.springframework.aop.SpringProxy is not visible from class loader
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:527)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:456)
	at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:294)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:225)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:291)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:193)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:585)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:913)
	at org.eclipse.gemini.blueprint.context.support.AbstractDelegatedExecutionApplicationContext.access$1600(AbstractDelegatedExecutionApplicationContext.java:60)
	at org.eclipse.gemini.blueprint.context.support.AbstractDelegatedExecutionApplicationContext$4.run(AbstractDelegatedExecutionApplicationContext.java:325)
	at org.eclipse.gemini.blueprint.util.internal.PrivilegedUtils.executeWithCustomTCCL(PrivilegedUtils.java:85)
	at org.eclipse.gemini.blueprint.context.support.AbstractDelegatedExecutionApplicationContext.completeRefresh(AbstractDelegatedExecutionApplicationContext.java:290)
	at org.eclipse.gemini.blueprint.extender.internal.dependencies.startup.DependencyWaiterApplicationContextExecutor$CompleteRefreshTask.run(DependencyWaiterApplicationContextExecutor.java:137)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.IllegalArgumentException: interface org.springframework.aop.SpringProxy is not visible from class loader
	at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:616)
	at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:592)
	at java.lang.reflect.WeakCache$Factory.get(WeakCache.java:244)
	at java.lang.reflect.WeakCache.get(WeakCache.java:141)
	at java.lang.reflect.Proxy.getProxyClass0(Proxy.java:455)
	at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:738)
	at org.springframework.aop.framework.JdkDynamicAopProxy.getProxy(JdkDynamicAopProxy.java:117)
	at org.springframework.aop.framework.ProxyFactory.getProxy(ProxyFactory.java:112)
	at org.springframework.aop.framework.autoproxy.AbstractAutoProxyCreator.createProxy(AbstractAutoProxyCreator.java:476)
	at org.springframework.aop.framework.autoproxy.AbstractAutoProxyCreator.wrapIfNecessary(AbstractAutoProxyCreator.java:362)
	at org.springframework.aop.framework.autoproxy.AbstractAutoProxyCreator.postProcessAfterInitialization(AbstractAutoProxyCreator.java:322)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsAfterInitialization(AbstractAutowireCapableBeanFactory.java:407)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1461)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:519)
	... 13 more
```

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionServiceImpl.java
larubbio: I don't see the distinction between domain and non-domain.  Also what non-domain entities would we cache?  It seems like we should have a single cache per "thing"

larubbio: Just to expand on this, what we are doing with this cache is creating a shortcut to find an instance of an object.  That object can be identified by many different "keys".  If we split our cache by key then we eliminate the caches ability to better manage the objects.  The cache internals may only maintain an object once even though multiple keys point to it.  However if we have many caches for the same object we prevent it from doing that for us.  (if the cache implementation actually does that)

frankhuster:
Now using single cache (per return type) and broadcasting cache evict to other nodes. Merging later this morning.

PR #591, opened 2015-09-21
Fixes NIP-160

larubbio:
For FLW services if we have no state for the user, look for deployed service in all the states in the
circle

(I'm relying on CI to run the tests for me.  I'm a bad dev committing without running them)


api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -234,6 +234,29 @@ private UserResponse getKilkariResponseUser(Long callingNumber) {
         return kilkariUserResponse;
     }
 
+    private boolean serviceDeplyedInCircle(Service service, Circle circle) {
```
frankhuster: Deplyed > Deployed

frankhuster:
Probably want to fix the typo, but otherwise - LGTM

larubbio:
I should probably add a test for it as well.

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -234,6 +234,29 @@ private UserResponse getKilkariResponseUser(Long callingNumber) {
...
+        }
+
+        List<State> stateList = circle.getStates();
+        if (stateList == null || stateList.isEmpty()) { // No state available
+            return true;
+        }
+
+        if (stateList.size() == 1) { // only one state available
+            return serviceDeployedInUserState(service, stateList.get(0));
+        }
```
koshalt: this if check could be collapsed into the for loop below right?

larubbio: Maybe.  If so the same should be done in the kilkari version of the code.
In fact maybe I should refactor them.

testing/src/test/java/org/motechproject/nms/testing/it/api/UserControllerBundleIT.java
```diff
@@ -3962,6 +3969,52 @@ public void verifyFT438() throws IOException, InterruptedException {
                 .getStatusCode());
     }
 
+    // Verify if a circle has multiple states and the service is not deployed in any of them than the call
```
frankhuster: then

frankhuster:
lgtm

PR #600, opened 2015-09-29
Static LMP & DOB options for test MCTS data

frankhuster:
Calling
```
python ~/src/mim/performance/data/data.py --cleardb --mctsmoms=10000 --lmp
```
with the `--lmp` parameter will generate mothers with a static LMP of 90 days prior to today.
And calling 
```
python ~/src/mim/performance/data/data.py --cleardb --mctskids=10000 --dob
```
with the `--dob` parameter will generate kids with a static DOB of today

### DIFF
--- a/testing/src/main/java/org/motechproject/nms/testing/web/TestingController.java
+++ b/testing/src/main/java/org/motechproject/nms/testing/web/TestingController.java
@@ -62,15 +62,21 @@ public class TestingController {
 
     @RequestMapping(value = "/createMctsMoms")
     @ResponseBody
-    public String createMctsMoms(@RequestParam(required = true) int count) throws IOException {
-        return testingService.createMctsMoms(count);
+    public String createMctsMoms(
+            @RequestParam(required = true) int count,
+            @RequestParam(required = true) boolean lmp)
+            throws IOException {
+        return testingService.createMctsMoms(count, lmp);
     }
 
 
     @RequestMapping(value = "/createMctsKids")
     @ResponseBody
-    public String createMctsKids(@RequestParam(required = true) int count) throws IOException {
-        return testingService.createMctsKids(count);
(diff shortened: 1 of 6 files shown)


testing/src/main/java/org/motechproject/nms/testing/web/TestingController.java
```diff
@@ -62,15 +62,21 @@ public String createSubscriptionPacks() {
 
     @RequestMapping(value = "/createMctsMoms")
     @ResponseBody
-    public String createMctsMoms(@RequestParam(required = true) int count) throws IOException {
-        return testingService.createMctsMoms(count);
+    public String createMctsMoms(
+            @RequestParam(required = true) int count,
```
koshalt: might be a silly question but how does this work for optional params?

frankhuster: this param isn't optional 
but the --lmp argument to data.py is
:-)

koshalt: right, and looking at data.py from above - 
`if args.mctsmoms > 0:
-        exec_http_get("{}/module/testing/createMctsMoms".format(args.server), {'count': args.mctsmoms})		+        params = {'count': args.mctsmoms, 'lmp': args.lmp > 0}
+        exec_http_get("{}/module/testing/createMctsMoms".format(args.server), params)`

How does this work? What kind of trickery is this? :confounded:

frankhuster: kosh == :cold_sweat: right now, but after reading [this](http://stackoverflow.com/questions/394809/does-python-have-a-ternary-conditional-operator), kosh == :sunglasses:

koshalt:
lgtm

PR #608, opened 2015-10-05
Fixes #607

### LINKED ISSUE #607: Child name should be optional in MCTS Import
From a mail from Pankaj:

> Just wanted to check if the Child name is mandatory while uploading the data in Motech,
> 
> If yes, then we need to change and make the child name optional as in India people don’t give names to baby immediately and it may be take even a couple of months to finalize the name.

larubbio:
Make child name optional in mcts child import


frankhuster:
Maybe add a test for it?

larubbio:
There was no test for it being required.  Also it doesn't look like we ever
access it, so I'm ok without a test.

PR #609, opened 2015-10-05
Obscuring msisdn from csv exception log

koshalt:
Fixes NIP-167

### DIFF
--- a/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
@@ -52,6 +52,7 @@ public class FrontLineWorkerImportServiceImpl implements FrontLineWorkerImportSe
     private static final String SUBCENTRE = "SubCentre_ID";
     private static final String CENSUS_VILLAGE = "Village_ID";
     private static final String NON_CENSUS_VILAGE = "SVID";
+    private static final int MASK_LENGTH = 3;
 
 
     private FrontLineWorkerService frontLineWorkerService;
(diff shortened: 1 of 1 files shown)


flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
```diff
@@ -131,13 +132,26 @@ private FrontLineWorker flwFromRecord(Map<String, Object> record, State state) {
...
                 throw new CsvImportDataException(String.format("Existing FLW with same MSISDN (%s) but " +
-                                        "different MCTS ID (%s != %s)", msisdn, mctsFlwId, flw.getMctsFlwId()));
+                                        "different MCTS ID (%s != %s)", obscureNumber(msisdn), mctsFlwId, flw.getMctsFlwId()));
             }
         }
 
         return flw;
     }
 
+    // obscure code from LogHelper in API package. Copy-paste since we cannot do a circular reference
```
frankhuster: What if you move LogHelper to the flw module?

koshalt: It is used more extensively in api than flw and that would be a stop gap anyways, without 'common' module that other modules can use.

koshalt: This felt like the best of all stop gaps

frankhuster: Okey dokey, LGTM!

On Mon, Oct 5, 2015 at 4:44 PM, kosh <notifications@github.com> wrote:

> In
> flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
> <https://github.com/motech-implementations/mim/pull/609#discussion_r41211913>
> :
>
> >              }
> >          }
> >
> >          return flw;
> >      }
> >
> > +    // obscure code from LogHelper in API package. Copy-paste since we cannot do a circular reference
>
> This felt like the best of all stop gaps
>
> —
> <https://github.com/motech-implementations/mim/pull/609/files#r41211913>.
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

PR #610, opened 2015-10-07
Adding sms reference number for MA reporting

koshalt:
Adding sms reference number for reporting


frankhuster:
How about adding a test that uses `smsReferenceNumber`?

koshalt:
Why? is it because you don't trust me?

koshalt:
How about I use it in a sentence instead? 
"I added code for smsReference" .

frankhuster:
LGTM

PR #611, opened 2015-10-08
Fixes NIP-166

larubbio:
The code wasn't looking to see if mcts id was set or not before checking to see if it matched the
id provided in the csv.


koshalt:
LGTM

PR #622, opened 2015-10-13
Record the type of the flw if it is provided in the mcts upload.

larubbio:


koshalt:
Lgtm. We don't want to enum the types and reject the unknown? Not that we use the types anywhere in code, but should be maybe upper/lower case it to standardize?

PR #625, opened 2015-10-14
Fixed mistaken reschedule of second weekly message

frankhuster:
Great bug find by Manjot (or  was is Rahul?)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,6 +156,11 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
     }
 
 
+    private boolean isFirstMessageOfWeek(String weekId) {
+        return weekId.endsWith("1");
```
larubbio: Can you add a comment?  Why does a weekId ending in 1 mean it is the first message of the week?

larubbio:
What was the bug that was found?  It's hard to tell if this change fixes it without knowing what it is.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,31 +156,62 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
...
+
+            // This message was never retried for this week before, so this is a first retry
+            LOGGER.debug(String.format("Creating retry for subscription %s and weekId %s",
+                    subscription.getSubscriptionId(), record.getWeekId()));
+
+            // Looking at the weekId, we can figure out if this was the first or second message for that week
+            // and calculate which day of the week we should reschedule the call for
+            DayOfTheWeek nextDay;
+            if (isFirstMessageOfWeek(record.getWeekId())) {
+                nextDay = subscription.getFirstMessageDayOfWeek().nextDay();
```
koshalt: The nextDay is irrelevant now right? The only reason we care about first vs second message/week is to set/update the retry call stage?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,31 +156,62 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
...
+                        getLanguageCode(subscription),
+                        getCircleName(subscription),
+                        subscription.getOrigin()
+                );
+                callRetryDataService.create(newCallRetry);
+            } else {
+                // this call was retried for a different week but never removed from the CallRetry table which means we
+                // never got a CDR from IMI, so let's warn about this and still try to reschedule it for this week.
+
+                Long id = (Long) callRetryDataService.getDetachedField(callRetry, "id");
```
koshalt: Shouldn't need this either since the retry extends MdsEntity. getId() should work

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,31 +156,62 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
...
+                        nextDay,
+                        CallStage.RETRY_1,
+                        record.getContentFileName(),
+                        record.getWeekId(),
+                        getLanguageCode(subscription),
+                        getCircleName(subscription),
+                        subscription.getOrigin()
+                );
+                callRetryDataService.create(newCallRetry);
+            } else {
```
koshalt: I'm confused by the else block here. This means that a retry entry already exists. Why would we overwrite it to Retry_1 again?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,31 +156,62 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
...
+                // this call was retried for a different week but never removed from the CallRetry table which means we
+                // never got a CDR from IMI, so let's warn about this and still try to reschedule it for this week.
+
+                Long id = (Long) callRetryDataService.getDetachedField(callRetry, "id");
+                LOGGER.warn(String.format("CallRetry record (id %d) for subscription %s for weekId %s was never " +
```
koshalt: I think that is serious enough to raise an alert at this point

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -156,31 +156,62 @@ private boolean isMsisdnInvalid(CallSummaryRecord record) {
...
-                    nextDay,
-                    CallStage.RETRY_1,
-                    record.getContentFileName(),
-                    record.getWeekId(),
-                    getLanguageCode(subscription),
-                    getCircleName(subscription),
-                    subscription.getOrigin()
-            );
-            callRetryDataService.create(newCallRetry);
+        if (callRetry == null || !callRetry.getWeekId().equals(record.getWeekId())) {
```
koshalt: This might be bordering on a nitpick but this might be a cleaner flow -

        if (callRetry == null) {
            // no retry was ever tried

            CallRetry newCallRetry = new CallRetry(
                    subscription.getSubscriptionId(),
                    msisdn,
                    null,
                    CallStage.RETRY_1,
                    record.getContentFileName(),
                    record.getWeekId(),
                    getLanguageCode(subscription),
                    getCircleName(subscription),
                    subscription.getOrigin()
            );
            callRetryDataService.create(newCallRetry);
            return;
        }

        if (!callRetry.getWeekId().equals(record.getWeekId())) {

            // this call was retried for a different week but never removed from the CallRetry table which means we
            // never got a CDR from IMI, so let's warn about this and still try to reschedule it for this week.
            Long id = callRetry.getId();
            LOGGER.warn(String.format("CallRetry record (id %d) for subscription %s for weekId %s was never " +
                            "deleted, which means we never received a CDR about it. Someone should look into this.", id,
                    callRetry.getSubscriptionId(), callRetry.getWeekId()));
            callRetry.setCallStage(CallStage.RETRY_1);
            callRetry.setWeekId(record.getWeekId());
            callRetryDataService.update(callRetry);
            return;
        }

        // We've already rescheduled this call, let's see if it needs to be re-rescheduled
        if ((subscription.getSubscriptionPack().retryCount() == 1) ||
                (callRetry.getCallStage() == CallStage.RETRY_LAST)) {

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: It might have been a leftover from a prior week and at a different call stage

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: yup

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: yup

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: yep

frankhuster:
@larubbio no bug was posted that I think. But while talking to Manjot about CDR processing past midnight (which works fine) I got a hunch something was a bit wonky with the second message for the week retries and looked at the code and indeed there was a bit of an issue.

.gitignore
```diff
@@ -29,3 +29,5 @@ docs/env
 docs/source/packages.rst
 docs/source/org
 .tx/config
+testing-directory/
```
koshalt: haaaaaaa---laaaayyy---looooo---yaaaaaahhhhh

koshalt:
lgtm

PR #626, opened 2015-10-16
Improvements from NIP-183 investigation

frankhuster:
Added some useful instrumentation & ops improvements while investigating NIP-183.
And one important fix.


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -161,23 +164,35 @@ private void rescheduleCall(Subscription subscription, CallSummaryRecord record,
...
+
+            String action = String.format("Creating retry for subscription: %s", subscription.getSubscriptionId());
+            try {
+                // This message was never retried before, so this is a first retry
+                LOGGER.debug(action);
+
+                CallRetry newCallRetry = new CallRetry(
+                        subscription.getSubscriptionId(),
+                        msisdn,
+                        DayOfTheWeek.fromInt(subscription.getStartDate().dayOfWeek().get()).nextDay(),
```
koshalt: is day of the week making a comeback?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/CallRetry.java
```diff
@@ -57,6 +66,7 @@ public CallRetry(String subscriptionId, Long msisdn, //NO CHECKSTYLE More than 7
         this.languageLocationCode = languageLocationCode;
         this.circle = circle;
         this.subscriptionOrigin = subscriptionOrigin;
+        this.timestamp = timestamp;
```
koshalt: if we are adding this field, maybe we should also just remove the dayOfTheWeek right now?

kilkari/src/main/java/org/motechproject/nms/kilkari/domain/CallRetry.java
```diff
@@ -43,11 +43,20 @@
     @Field
     private SubscriptionOrigin subscriptionOrigin;
 
+    /**
+     * the time at which this callRetry was created (extracted from a RequestId)
+     * allows us to check for duplicate callRetry and end up with better CDR processing logic when dealing with
+     * bad data from IMI
+     */
+    @Field
+    private String timestamp;
```
koshalt: why not just store the whole request id instead of parsing it for a timestamp string?

koshalt:
test this please

PR #636, opened 2015-10-29
Fixes #628, Fixes #629, fixes #630, fixes #631, fixes #649

### LINKED ISSUE #628: Region Handling 2.0
This ticket supersedes the Taluka <> HealthBlock mapping ticket ( #595 ) 

The new plan for managing the import of data below Taluka as well as the constantly changing location hierarchy is as follows:

1. Decouple the validation checks ( #629 )
2. Create missing locations on beneficiary and FLW upload ( #630 )
3. Upsert locations when the location file is uploaded ( #631 )
4. Find HealthBlocks using state not taluka id ( #649 )

### LINKED ISSUE #629: Decouple Location Validation Checks
When validating a location we should validate the location entities (State, District, Taluka, Village) separately from the Health entities (HealthBlock, HealthFacility, HealthSubFacility)

A HealthBlock will still maintain a relationship to a taluka, but that taluka may not be the taluka a user is located in.  In the future if required we can import the taluka<>HealthBlock mapping in which case we could in theory reinstate this validation.

We are building this with the assumption that a HealhBlock will never span state boundaries.

### LINKED ISSUE #630: Create missing locations on beneficiary and FLW upload
When uploading mother, child and FLW files (or while importing via web services) if the location for the user does not exist we should create new location entities for them.  This may require relaxing the not-null constraints on those entities.

We also need to create missing locations when importing the language file.

### LINKED ISSUE #631: Upsert locations when the location file is uploaded
When importing updated location files, the import should first attempt to find the location entity by it's id, create it if it isn't found and update it if it is.

### LINKED ISSUE #649: healthBlockService.findByTalukaAndCode needs to match on state not taluka
Since Taluka <-> HealthBlock is many to many we can't look up a health block using taluka id.

If we first create a taluka from the MCTS location files then we will associate the health block with the taluka in that file, however if we first see a new health block from a beneficiary or flw upload then we will use that taluka id.  Additionally the second time we see a healthblock it may be associated with a different id. 

So instead of looking up based on taluka id, we have to find the state code of the given taluka and match it based on the state code of the taluka associated with the health block

frankhuster:
When uploading or updating MCTS FWL or beneficiaries, will create sub-district location data if it's not found in the database. Additionally updated some location data assumptions we made about field length and nullability.

IMPORTANT NOTE: this likely requires a "by hand" schema update, since we changed some field lengths. So we'll have to deliver a SQL script for OPS to run during the deployment.


koshalt:
Look good, can you also add a test for beneficiary import?

region/src/main/java/org/motechproject/nms/region/domain/HealthFacility.java
```diff
@@ -44,8 +43,6 @@
     private Long code;
 
     @Field
-    @Column(allowsNull = "false")
-    @NotNull
     private HealthFacilityType healthFacilityType;
```
larubbio: HealthFacilityType won't come in the mcts files right?  If so are we ever going to set it?  I think it might be needed in reports

region/src/main/java/org/motechproject/nms/region/domain/HealthFacility.java
```diff
@@ -142,8 +139,6 @@ public String toString() {
         return "HealthFacility{" +
                 "name='" + name + '\'' +
                 ", code=" + code +
-                ", healthFacilityType=" + healthFacilityType +
```
larubbio: Any reason why these were removed?

region/src/main/java/org/motechproject/nms/region/domain/HealthSubFacility.java
```diff
@@ -108,7 +107,6 @@ public String toString() {
         return "HealthSubFacility{" +
                 "name='" + name + '\'' +
                 ", code=" + code +
-                ", healthFacility=" + healthFacility +
```
larubbio: Any reason why this was removed?

region/src/main/java/org/motechproject/nms/region/domain/Taluka.java
```diff
@@ -160,7 +157,6 @@ public String toString() {
         return "Taluka{" +
                 "name='" + name + '\'' +
                 ", code=" + code +
-                ", district=" + district +
```
larubbio: Any reason why this was removed?

region/src/main/java/org/motechproject/nms/region/service/impl/LocationServiceImpl.java
```diff
@@ -66,89 +75,131 @@ public LocationServiceImpl(StateDataService stateDataService, DistrictService di
         this.healthSubFacilityService = healthSubFacilityService;
     }
 
+
+    private boolean isValidID(final Map<String, Object> map, final String key) {
```
larubbio: A valid id is a long > than 0 or any string other than "0"?  Does this need to be a more involved check?  Can an Integer ever come in?  Should we use a regex that verifies it only contains 0-9.  Or is there no need to do those extra checks.

region/src/main/java/org/motechproject/nms/region/service/impl/LocationServiceImpl.java
```diff
@@ -28,14 +30,21 @@
 @Service("locationService")
 public class LocationServiceImpl implements LocationService {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(LocationServiceImpl.class);
+
     private static final String INVALID = "<%s - %s : Invalid location>";
     private static final String STATE = "StateID";
```
larubbio: Since the other constants were changed to be _ID should the same be done for STATE and DISTRICT?

region/src/main/java/org/motechproject/nms/region/service/impl/VillageServiceImpl.java
```diff
@@ -47,12 +47,14 @@ public Village execute(Query query) {
     }
 
     @Override
-    public void create(Village village) {
-        dataService.create(village);
+    public Village create(Village village) {
+        Village v = dataService.create(village);
+        dataService.evictEntityCache(true);
```
larubbio: Why does village evict the cache but none of the other location entities evict the cache?

larubbio:
I think this fixes the first issue since it creates new locations if the ones in the file don't exist.  I don't think it fixes the validation piece.

Assume we have Taluka A & D and HealthBlocks B & C.  In the database B is associated with A.  C is associated with Taluka A and D in the database (but also with A in reality.)   In this case the code in https://github.com/motech-implementations/mim/blob/629/region/src/main/java/org/motechproject/nms/region/domain/validation/FullLocationValidator.java#L157 will fail.  I think we also need to remove that check to fix #629

larubbio:
Does this actually fix #631?  It doesn't upsert the locations.  I think it fixes #630 instead.

region/src/main/java/org/motechproject/nms/region/service/impl/LocationServiceImpl.java
larubbio: done

frankhuster:
So how are we going to handle the data migration?

frankhuster:
I guess we'll have to add some sort of an ALTER TABLE SQL script in a separate PR.

PR #639, opened 2015-10-29
fixes NIP-192

larubbio:
This commit moves the relationship of circles down from state to district.  When the language csv is uploaded if a circle is not found it is created.  Also when we import users we use district to find their circle.


koshalt:
this build is full of lies

koshalt:
I wouldn't merge this just yet. Pretty unreliable runs locally and something fishy about the district collection move from list to sets in States.

koshalt:
http://stackoverflow.com/questions/9619955/jdo-duplicate-entry-exception 

This might have some clues about reassociating detached objects. It still doesn't explain by just running each one of the 3 tests by themselves passes while running the whole test class fails. All that I have had from the last 4hrs is symptoms :cry:

larubbio:
ok, so I wasn't just delirious from jetlag.  On the plan I was wondering if
it was the changes in the region helper class.  Now when you create a
circle it creates the district (which calls the circle create).  That
should be ok, but if you first create the district it calls the circle
create which calls create the district which could create two of them.
Although if that was the case then it wouldn't fail when run alone.  Or
maybe it would, I don't know.

koshalt:
Ok, the tests should be fixed now.. except I have no logical explanation for why/how they got fixed. I moved the subscription creation logic away from the main test function and reused makeMctsSubscription(...) like some other tests do and it works now.

The helper does exactly what the main function used to do, so this is still a mystery...but I'm guess that this might be more of an infrastructure (or cleanup) problem than a real one.

frankhuster:
Hey @larubbio, @koshalt is out, want me to review that or are you still working on it?

region/src/main/java/org/motechproject/nms/region/domain/State.java
```diff
@@ -80,22 +68,14 @@ public void setCode(Long code) {
         this.code = code;
     }
 
-    public List<District> getDistricts() {
+    public Set<District> getDistricts() {
```
koshalt: does changing to set from list mean a schema migration?

koshalt:
LGTM. @frankhuster it might be good for you to look at this as well since I made some changes in this PR as well. 2 questions - 
1. does this need to go into 1.0.3?
2. does list -> set = schema change?

region/src/main/java/org/motechproject/nms/region/domain/State.java
larubbio: With a list there is a column added on the many side that is no longer used when the variable is a set.  So ideally that column would be dropped, but the schema should be compatible.

larubbio:
@frankhuster If you are logging in and wouldn't mind reviewing this, go for it.  Although I also get that you are on vacation at the beach.  I feel dirty even asking.

region/src/main/java/org/motechproject/nms/region/domain/State.java
frankhuster: Since we're going to have to actually change the database (for the location schema change) maybe we could get rid of the column?

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/MctsBeneficiaryImportServiceBundleIT.java
```diff
@@ -1313,7 +1319,30 @@ public void testCreateMotherSubscriptionWhenDeactivatedChildSubscriptionExists()
         subscriber = subscriberDataService.findByNumber(9439986187L);
         assertEquals(1, subscriber.getActiveAndPendingSubscriptions().size());
         assertEquals(2, subscriber.getAllSubscriptions().size());
-        
+
+    }
+
+    /*
+     * To verify pregnancy record gets created when child record exist with status as deactivated.
```
frankhuster: The comment doesn't match the test name, right?

larubbio: Nope.  I probably copied another test and grabbed the comment by mistake.

frankhuster:
Minor copy/paste comment crap, otherwise LGTM

frankhuster:
LGTM

PR #642, opened 2015-10-30
Small fixes from NIP-198 investigation

frankhuster:
failureReasons => failureReason   (no trailing s)
more verbose log output on error
checksums now case insensitive


koshalt:
LGTM

PR #643, opened 2015-10-31
Fixes NIP-202

frankhuster:
Invalid subscriptions in phase2 of CDR processing should not cause
an overall rejection, but instead just send an alert & log

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
@@ -1,5 +1,6 @@
 package org.motechproject.nms.imi.service.impl;
 
+import org.apache.commons.lang3.StringUtils;
 import org.motechproject.alerts.contract.AlertService;
 import org.motechproject.alerts.domain.AlertStatus;
 import org.motechproject.alerts.domain.AlertType;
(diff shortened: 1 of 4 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/CsrValidatorServiceImpl.java
```diff
@@ -135,15 +142,26 @@ private void validateLanguageLocationCode(String code) {
...
+        }
+        return sub;
+    }
+
+
     //todo: IT
     @Override
     public void validateSummaryRecord(CallSummaryRecordDto record) {
-        Subscription sub = subscriptionService.getSubscription(record.getRequestId().getSubscriptionId());
+        Subscription sub = validateSubscription(record.getRequestId().getSubscriptionId());
```
koshalt: can we rename this to validateSummaryRecordDto? The overload on the type seems a little confusing.

koshalt:
LGTM, other than the renaming. Are there any downstream/upstream impacts to working with a null subscription post the validation?

frankhuster:
Will rename, and can't think of any impacts any worse than outright
rejection right now

On Fri, Oct 30, 2015 at 8:41 PM, kosh <notifications@github.com> wrote:

> LGTM, other than the renaming. Are there any downstream/upstream impacts
> to working with a null subscription post the validation?
>
> —
> <https://github.com/motech-implementations/mim/pull/643#issuecomment-152690031>
> .
>



-- 
Frank Huster
Mobile Health Innovations
Grameen Foundation
+1.206.325.6690 x201

PR #652, opened 2015-11-09
Faster CDR processing

frankhuster:
Fixes NIP-201
Does not verify all entities in CDR/CSR in phase 2 anymore, delegates that to Phases 3 & 4
More & better debug & info logging.


imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -12,6 +12,7 @@
 import org.motechproject.event.listener.EventRelay;
 import org.motechproject.event.listener.annotations.MotechListener;
 import org.motechproject.mds.query.SqlQueryExecution;
+import org.motechproject.metrics.service.Timer;
```
larubbio: I think we might have two Timer like classes now.  org.apache.commons.lang.time.StopWatch was introduced during the MCTS web service work.  Ideally we would consolidate, I'm not sure if the time it takes to do that is worth it now.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -608,7 +632,7 @@ private File localCdrDir() {
     @Override
     public void verifyDetailFileChecksumAndCount(CdrFileNotificationRequest request) {
 
-        LOGGER.debug("Starting CDR Process Phase 1");
+        LOGGER.info("Phase 1 - Start");
```
larubbio: Any reason why you removed "CDR Process"?  Seems like a useful string to search for

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -707,58 +730,89 @@ public void verifyDetailFileChecksumAndCount(CdrFileNotificationRequest request)
...
+        LOGGER.info("Phase 4 - Start");
 
         // Phase 4: distribute the processing of CSRs to all nodes, make backup copy of all received CSR
         File csrFile = new File(localCdrDir(), request.getCdrSummary().getCdrFile());
+        LOGGER.info("Phase 4 - sendSummaryRecords - distributing CSRs");
         errors = sendSummaryRecords(csrFile);
         if (errors.size() > 0) {
-            reportAuditAndPost(request.getCdrSummary().getCdrFile(), errors);
+            reportAuditAndPost(request.getFileName(), errors);
+            LOGGER.debug(CDR_PROCESS_PHASE_2_ERROR);
```
larubbio: Is this a phase 2 error or an error in phase 2, 3 or 4?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -320,6 +326,29 @@ private CallSummaryRecord aggregateSummaryRecord(CallSummaryRecordDto csr) {
     }
 
 
+    private void validateCallSummaryRecord(Subscription subscription, CallSummaryRecordDto csr) throws InvalidCdrData {
+        if (!subscription.getSubscriptionPack().hasMessageWithWeekId(csr.getWeekId())) {
+            throw new InvalidCdrData(String.format("%s is an invalid weekId for pack %s",
+                    csr.getWeekId(), subscription.getSubscriptionPack().getName()));
+        }
+
+        if (!subscription.getSubscriptionPack().hasMessageWithFilename(csr.getContentFileName())) {
```
larubbio: Is there a race condition with updates to the content names?  Just wondering if they ever release new content will this fail?  Not sure it is a likely scenario.  Mostly just curious

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
```diff
@@ -320,6 +326,29 @@ private CallSummaryRecord aggregateSummaryRecord(CallSummaryRecordDto csr) {
...
+            throw new InvalidCdrData(String.format("%s is an invalid contentFilename for pack %s",
+                    csr.getContentFileName(), subscription.getSubscriptionPack().getName()));
+        }
+
+        if (circleService.getByName(csr.getCircle()) == null) {
```
larubbio: Don't we tell them what the circle is and they just return it back?  If we don't get a circle do we still want to record the call?

larubbio:
LGTM.  Some small questions/comments.  Look them over when you get a chance and then feel free to merge.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: I would expect the content filename to be updated in the pack beforehand - if not then this error would be warranted. We could also not verify, but if feels a bit weird.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrServiceImpl.java
frankhuster: Hmm, given the essence of this check-in was to make things faster, Should I remove these checks altogether? It seems to be much faster now, so I'm thinking of leaving them, but I guess they should be prime chop candidates.

PR #654, opened 2015-11-15
Better CDR processing

frankhuster:
More permissive of errors
More distributed tasks
Idempotent (gotta check that a bit more)

Still has that hideous performance issue when overloaded with messages, really annoying and weird.


koshalt:
ok to test

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -259,54 +202,47 @@ private void sendProcessSummaryRecordEvent(CallSummaryRecordDto record) {
...
-                fis.close();
 
-                thisChecksum = ChecksumHelper.checksum(file);
+                lineNumber++;
 
-            } catch (IllegalArgumentException e) {
-                String error = String.format("Unable to read header %s: %s", fileName, e.getMessage());
-                errors.add(error);
             }
+            reader.close();
```
koshalt: possibly move this to try-finally or using resource

frankhuster: But, am I not using one above?

PR #655, opened 2015-11-19
v1.0.4 - Distributed & Local CSR processing

frankhuster:
Local vs. distributed CSR processing can now be controlled by the following imi.properties boolean property setting:
```
imi.distributed_csr_processing
```
The default value is `false`.


koshalt:
Is the config addition not part of this change? intentional?

PR #656, opened 2015-11-24
CDR chunking

frankhuster:
A new imi.properties setting enables chunk CSR processing:
```
imi.csr_chunk_size
```
Its default value is 1, which means no chunking.
Any other value will distribute CSR in chunks for processing.

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/domain/ChunkAuditRecord.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/domain/ChunkAuditRecord.java
@@ -1,0 +1,84 @@
+package org.motechproject.nms.imi.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+import javax.jdo.annotations.Column;
+import javax.jdo.annotations.Unique;
+
+/**
+ * CallSummaryRecord Processing Audit - records CSR chunk processing information
+ */
+@Entity(tableName = "nms_imi_chunk_audit_record")
+@Unique(name = "UNIQUE_CHUNK_AUDIT_RECORD_COMPOSITE_IDX", members = { "file", "chunk" })
+public class ChunkAuditRecord {
+
+    @Field
+    @Column(allowsNull = "false")
+    private String file;
+
+    @Field
+    @Column(allowsNull = "false")
+    private String chunk;
+
+    @Field
+    @Column(allowsNull = "false")
+    private int csrToProcess;
+
+    @Field
+    private int csrProcessed;
+
+    @Field
+    private String node;
+
+    public ChunkAuditRecord() {
+    }
+
+    public ChunkAuditRecord(String file, String chunk, int csrToProcess, int csrProcessed, String node) {
+        this.file = file;
+        this.chunk = chunk;
+        this.csrToProcess = csrToProcess;
+        this.csrProcessed = csrProcessed;
+        this.node = node;
+    }
+
+    public String getFile() {
+        return file;
+    }
+
+    public void setFile(String file) {
+        this.file = file;
+    }
+
+    public String getChunk() {
+        return chunk;
+    }
+
+    public void setChunk(String chunk) {
+        this.chunk = chunk;
+    }
+
+    public int getCsrToProcess() {
+        return csrToProcess;
+    }
+
+    public void setCsrToProcess(int csrToProcess) {
+        this.csrToProcess = csrToProcess;
+    }
+
+    public int getCsrProcessed() {
+        return csrProcessed;
+    }
+
+    public void setCsrProcessed(int csrProcessed) {
+        this.csrProcessed = csrProcessed;
+    }
+
+    public String getNode() {
+        return node;
+    }
+
+    public void setNode(String node) {
+        this.node = node;
+    }
+}
--- a/imi/src/main/java/org/motechproject/nms/imi/domain/ChunkAuditRecord.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/domain/ChunkAuditRecord.java
@@ -1,5 +1,6 @@
 package org.motechproject.nms.imi.domain;
 
+import org.joda.time.DateTime;
 import org.motechproject.mds.annotations.Entity;
 import org.motechproject.mds.annotations.Field;
 
@@ -9,7 +10,7 @@ import javax.jdo.annotations.Unique;
 /**
  * CallSummaryRecord Processing Audit - records CSR chunk processing information
  */
-@Entity(tableName = "nms_imi_chunk_audit_record")
+@Entity(tableName = "nms_imi_chunk_audit_records")
 @Unique(name = "UNIQUE_CHUNK_AUDIT_RECORD_COMPOSITE_IDX", members = { "file", "chunk" })
 public class ChunkAuditRecord {
 
@@ -31,15 +32,22 @@ public class ChunkAuditRecord {
     @Field
     private String node;
 
+    @Field
+    private DateTime processingStart;
+
+    @Field
+    private DateTime processingEnd;
+
+    @Field
+    private String timing;
+
     public ChunkAuditRecord() {
     }
 
-    public ChunkAuditRecord(String file, String chunk, int csrToProcess, int csrProcessed, String node) {
+    public ChunkAuditRecord(String file, String chunk, int csrToProcess) {
         this.file = file;
         this.chunk = chunk;
         this.csrToProcess = csrToProcess;
-        this.csrProcessed = csrProcessed;
-        this.node = node;
     }
 
     public String getFile() {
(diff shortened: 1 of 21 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +373,129 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+        Map<String, Object> params = new HashMap<>();
+        params.put("file", file);
+        params.put("name", name);
+        params.put("chunk", chunk);
+        MotechEvent motechEvent = new MotechEvent(NMS_IMI_PROCESS_CHUNK, params);
+        eventRelay.sendEventMessage(motechEvent);
+    }
+
+
+    private String hostName() {
```
koshalt: nice. maybe we can move this to the post-contruct declaration eventually?

imi/src/main/java/org/motechproject/nms/imi/domain/ChunkAuditRecord.java
```diff
@@ -0,0 +1,84 @@
...
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+import javax.jdo.annotations.Column;
+import javax.jdo.annotations.Unique;
+
+/**
+ * CallSummaryRecord Processing Audit - records CSR chunk processing information
+ */
+@Entity(tableName = "nms_imi_chunk_audit_record")
```
koshalt: table name is plural right? :laughing:

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +373,129 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+
+
+    private void reportIfAllChunksWereProcessed(final String file) {
+
+        @SuppressWarnings("unchecked")
+        SqlQueryExecution<List<ChunkAuditRecord>> queryExecution = new SqlQueryExecution<List<ChunkAuditRecord>>() {
+
+            @Override
+            public String getSqlQuery() {
+                String query = String.format("SELECT * FROM nms_imi_chunk_audit_record WHERE file = '%s' AND " +
```
koshalt: need to update the table name here, fyi

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +373,129 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+            updateChunkAuditRecord(file, name, csrDtos.size());
+            reportIfAllChunksWereProcessed(file);
+
+        } catch (Exception e) {
+            String msg = String.format(MOTECH_BUG, "P5 - processChunk - " + name, ExceptionUtils.getFullStackTrace(e));
+            LOGGER.error(msg);
+            alertService.create(name, "processCsrs", msg.substring(0, min(msg.length(), MAX_CHAR_ALERT)),
+                    AlertType.CRITICAL, AlertStatus.NEW, 0, null);
+            // We want to fail this event to we get retried
+            throw e;
```
koshalt: shouldn't throw here right? The message keeps going back on the queue?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +373,129 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+        String name = (String) event.getParameters().get("name");
+        String json = (String) event.getParameters().get("chunk");
+        ObjectMapper mapper = new ObjectMapper();
+        List<CallSummaryRecordDto> csrDtos;
+
+        try {
+            try {
+                csrDtos = mapper.readValue(json, new TypeReference<List<CallSummaryRecordDto>>() { });
+            } catch (IOException e) {
+                throw new ChunkingException(String.format("Exception unpacking packaging CSR chunk %s: %s", name,
```
koshalt: If nobody catches this specific exception type, then just let the bigger try catch-all and log? we can get rid of this inner try-catch?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +373,129 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+        } catch (Exception e) {
+            String msg = String.format(MOTECH_BUG, "P5 - processChunk - " + name, ExceptionUtils.getFullStackTrace(e));
+            LOGGER.error(msg);
+            alertService.create(name, "processCsrs", msg.substring(0, min(msg.length(), MAX_CHAR_ALERT)),
+                    AlertType.CRITICAL, AlertStatus.NEW, 0, null);
+            // We want to fail this event to we get retried
+            throw e;
+        }
+
+        LOGGER.info("Processed {} {}", name, timer.frequency(csrDtos.size()));
```
koshalt: maybe we should have this logger in a finally block?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -392,8 +555,21 @@ public void processCsrs(File file) {
...
 
-                    processCsrEvent(csr.toDto(), distributedProcessing);
-                    processCount++;
+                    if (chunkSize > 1) {
+                        chunk.add(csr.toDto());
+                        if (chunk.size() >= chunkSize || lineNumber >= lineCount) {
+                            LOGGER.info(String.format("Distributing chunk %d of %d (%d csrs): %s", chunkNumber,
+                                    chunkCount, chunk.size(), chunkTimer.frequency(chunkNumber)));
+                            String chunkName = String.format("Chunk%d/%d", chunkNumber, chunkCount);
+                            distributeChunk(fileName, chunkName, chunk);
```
koshalt: distributeChunk can throw a chunking exception but no special catch for it here?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Well, just in case the exception was transient?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: yeah, that makes sense...

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -330,12 +361,24 @@ private boolean shouldDistributeCsrProcessing() {
...
         }
     }
 
 
-    private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
+    private int csrChunkSize() {
+        try {
+            return Integer.valueOf(settingsFacade.getProperty(CSR_CHUNK_SIZE));
+        } catch (Exception e) {
+            LOGGER.info("Unable to read {}, returning default value {}", CSR_CHUNK_SIZE, CSR_CHUNK_SIZE_DEFAULT);
```
koshalt: this would be warn or error instead of info?

frankhuster: You used debug in cleanOldCallRecords and I'm fine with that.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -346,6 +389,178 @@ private void processCsrEvent(CallSummaryRecordDto csrDto, boolean distributed) {
...
+            LOGGER.info("CSR processing complete - {} processed in {}.", file, reportChunkProcessingTime(file));
+        }
+    }
+
+
+    private void upsertChunkAuditRecord(String file, String chunk, int csrCount) {
+        ChunkAuditRecord record = chunkAuditRecordDataService.findByFileAndChunk(file, chunk);
+        if (record == null) {
+            chunkAuditRecordDataService.create(new ChunkAuditRecord(file, chunk, csrCount));
+        } else {
```
koshalt: ideally, we would never ever get into the else block right?

koshalt: unless the file gets run again?

koshalt:
lgtm. waiting for build to pass before merge

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: done

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: that's right

PR #657, opened 2015-11-28
Fixes NMS-300

frankhuster:
Also misc logging improvements.


imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -1109,6 +1113,8 @@ public void cdrProcessPhase5(MotechEvent event) {
             return;
         }
 
+        sendPhaseEvent(KK_CSR_VERIFIER_READ_DOMAIN_DATA, request);
```
larubbio: Why is this event sent?

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -718,7 +722,7 @@ private File localCdrDir() {
     //          while collecting a list of errors on the go.
     //          Does not proceed to phase 2 if any error occurred and returns an error
     @Override //NO CHECKSTYLE Cyclomatic Complexity
-    public void cdrProcessingPhase1(CdrFileNotificationRequest request) {
+    public void cdrProcessPhase1(CdrFileNotificationRequest request) {
```
larubbio: Is there a reason for the name change or just a random inclusion

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/CsrVerifierServiceImpl.java
```diff
@@ -32,32 +33,34 @@
...
-        circles = new HashSet<>();
-        for (Circle circle : circleService.getAll()) {
-            circles.add(circle.getName());
-        }
-        LOGGER.info("Loaded {} circles from database.", circles.size());
-
-        languages = new HashSet<>();
-        for (Language language : languageService.getAll()) {
-            languages.add(language.getCode());
+    private synchronized void readDomainData() {
```
larubbio: My preference would be to use spring cache as opposed to rolling our own.

larubbio:
No major complaints, would  prefer to use spring cache then our own, but otherwise lgtm.

frankhuster:
Argh, you're right. I'll change the code to use Spring cache.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: Random inclusion: made it consistent with all the other cdrProcessPhaseX methods

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -1109,6 +1113,8 @@ public void cdrProcessPhase5(MotechEvent event) {
             return;
         }
 
+        eventRelay.broadcastEventMessage(new MotechEvent(CSR_VERIFIER_CACHE_EVICT_MESSAGE));
```
koshalt: should we move this to the beginning of the method (before file copy) so that the cache evict message has guarateed time to run before processCsrs kicks off? 

Not that we should have implicit timing dependencies in code, but just a thought.

koshalt:
lgtm.

imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
frankhuster: No need, it takes quite a while to queue up all the CSR chunk dispatch events.
The message will be broadcast and processed way before the first dispatched chunk is being processed.

PR #663, opened 2015-12-16
Ivr subscription update

koshalt:
Fixes NIP-281

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
@@ -124,7 +124,7 @@ public class SubscriberServiceImpl implements SubscriberService {
 
     @Override
     @Transactional
-    public void update(Subscriber subscriber) {
+    public void updateStartDate(Subscriber subscriber) {
 
         Subscriber retrievedSubscriber = subscriberDataService.update(subscriber);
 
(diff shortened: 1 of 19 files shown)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -142,10 +142,11 @@ public void update(Subscriber subscriber) {
                     subscriptionService.updateStartDate(subscription, subscriber.getLastMenstrualPeriod());
                 }
 
-            } else if (subscription.getSubscriptionPack().getType() == SubscriptionPackType.CHILD) {
-
-                subscriptionService.updateStartDate(subscription, subscriber.getDateOfBirth());
+            } else { // SubscriptionPackType.CHILD
 
+                if (subscriber.getDateOfBirth() != null) {  // Subscribers via IVR will not have DOB
```
koshalt: btw, this is the fix with test added below.

frankhuster:
LGTM, maybe (in your IT) try creating the subscriber using the same http endpoint that is used when a subscriber is created via IVR?

koshalt:
Just took a look at the existing tests to create subscriber via http and we would have to share a bunch of stuff between the api and kilkari test modules. Also, test probably belongs more in kilkari since it is actually testing an update mechanism outside of the api layer. tl;dr; - too much effort for little gain and I'm lazy. Merging.

PR #666, opened 2015-12-23
Fixes NMS-301

frankhuster:
By ignoring subscriptions on their last day from fresh calls during generateTargetFile


koshalt:
LGTM

PR #670, opened 2016-01-05
Fixes NIP-317 - Activity count

koshalt:
Fixes NIP-317 with creation of new activity record for second attempt

### DIFF
--- a/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
+++ b/mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
@@ -243,7 +243,7 @@ public class CourseNotificationServiceImpl implements CourseNotificationService 
                     SMS_CONTENT_PREFIX + smsLanguageProperty);
         }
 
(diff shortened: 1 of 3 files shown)


mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/CourseNotificationServiceImpl.java
```diff
@@ -243,7 +243,7 @@ private String buildSmsContent(Long callingNumber) {
                     SMS_CONTENT_PREFIX + smsLanguageProperty);
         }
 
-        int attempts = activityService.getAllActivityForUser(callingNumber.toString()).size();
+        int attempts = activityService.getCompletedActivityForUser(callingNumber.toString()).size();
```
frankhuster: what's that for?

koshalt: this is for the attempt count in the completion sms. Back in the day, we only used to track a start activity and hence the old way was reliable. Along the way, we had to add another activity to track completion as well and this turned into a hidden bug since we would be double counting with 'all' activities

frankhuster:
LGTM

PR #674, opened 2016-02-01
Set defaultFetchGroup to false on CallContent member of FLW CDR entity

frankhuster:
Should make getCurrentMonthlyUsageForFLWAndService query faster


larubbio:
lgtm

PR #683, opened 2016-03-23
Block all existing anonymous MA users

koshalt:


koshalt:
@frankhuster all yours

api/src/main/java/org/motechproject/nms/api/web/UserController.java
```diff
@@ -256,13 +257,12 @@ private UserResponse getFrontLineWorkerResponseUser(String serviceName, Long cal
...
-        if (flw == null) {
-            // New requirement - https://applab.atlassian.net/projects/NMS/issues/NMS-325 - Block anonymous FLWs
-            // if flw is null here, we don't already have a record from MCTS. return 403
-            if (MOBILE_ACADEMY.equals(serviceName)) {
-                throw new NotAuthorizedException(String.format(NOT_AUTHORIZED, CALLING_NUMBER));
-            }
-        } else {
+        if (MOBILE_ACADEMY.equals(serviceName)) {
+            // make sure that flw is authorized to use MA
+            restrictAnonymousMAUserCheck(flw);
```
frankhuster: http://bit.ly/1VEzzNI ?

testing/src/test/java/org/motechproject/nms/testing/it/api/utils/ApiTestHelper.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.motechproject.nms.flw.domain.FrontLineWorker;
+import org.motechproject.nms.flw.domain.FrontLineWorkerStatus;
+
+/**
+ * Api test helper with static methods
+ */
+public final class ApiTestHelper {
+
+    public static FrontLineWorker createFlw(String name, Long phoneNumber, String mctsFlwId, FrontLineWorkerStatus status) {
```
frankhuster: If you're gonna do that, why not also add createFlwWithDistrictAndState and save yourself a little extra work?

testing/src/test/java/org/motechproject/nms/testing/it/api/UserControllerBundleIT.java
```diff
@@ -4785,13 +4789,45 @@ public void verifyFT331() throws IOException, InterruptedException {
...
+        deployedServiceDataService.create(new DeployedService(rh.delhiState(),
+                Service.MOBILE_ACADEMY));
+
+        // National Capping
+        ServiceUsageCap stateUsageCap = new ServiceUsageCap(null,
+                Service.MOBILE_KUNJI, 150);
+        serviceUsageCapDataService.create(stateUsageCap);
+
+        // FLW
+        FrontLineWorker flw = ApiTestHelper.createFlw("Claire Underwood", 1200000000l, null, FrontLineWorkerStatus.ACTIVE);
```
frankhuster: She's so _awesomely_cunning_ :smiling_imp:

frankhuster:
Couple of nitpicks you're welcome to ignore.
I see that you added verifyDenialWhenMctsFlwIdMissing and I assume there are tests in UserControllerBundleIT that verify the opposite (non-anonymous MCTS FLWs can go through) is working properly, right?

LGTM

api/src/main/java/org/motechproject/nms/api/web/UserController.java
llavoie: @frankhuster Which part? I don't think we can assume that the service is MA here -- could be MK (although if MK ever comes online we'll probably be asked to disable anonymous access there too...)

Or were you referring to the restrictAnonymousUserCheck? There's a case where the flw might be non-null but the user is ANONYMOUS (because we have created FLW records for non-MCTS FLWs who have started the course prior to these changes).

frankhuster: I was referring to the fact that the naming of the restrictAnonymousUserCheck method is bad grammar, it should be named either restrictAnonymousUser or checkRestrictedAnonymousUser.
But that's definitely http://bit.ly/1Zt2PHh

koshalt: trump

koshalt: noproblem

koshalt:
Yes, that is correct.

testing/src/test/java/org/motechproject/nms/testing/it/api/utils/ApiTestHelper.java
koshalt: I could have and probably should have... but don't want to go down that rabbit hole of trying to fix the tests the right way :disappointed:

PR #686, opened 2016-05-31
Disabling hpd check for MA mcts sync

koshalt:


koshalt:
ok to test

frankhuster:
Why about having HPD checking enabled or disabled based on a setting instead of just ripping out the code?

koshalt:
@frankhuster That's exactly what the ripped out code used to do...and it still does for kilkari. Turns out that flw sync was never meant to be part of the hpd check and that was lost in communication.

frankhuster:
Okeydokey, then LGTM

PR #688, opened 2016-06-06
Consume NIC dump

koshalt:
This implements an ops hook to consume NIC data from an http client with specific fields in a contract. This helps sync up one-time db dumps to make up for MCTS service availability (lack of)

### DIFF
--- a/api/src/main/java/org/motechproject/nms/api/web/OpsController.java
+++ b/api/src/main/java/org/motechproject/nms/api/web/OpsController.java
@@ -2,6 +2,8 @@ package org.motechproject.nms.api.web;
 
 import org.motechproject.event.MotechEvent;
 import org.motechproject.event.listener.EventRelay;
+import org.motechproject.nms.flw.service.FrontLineWorkerService;
+import org.motechproject.nms.flw.utils.FlwConstants;
 import org.motechproject.nms.imi.service.CdrFileService;
 import org.motechproject.nms.kilkari.repository.SubscriptionDataService;
 import org.motechproject.nms.kilkari.service.SubscriptionService;
@@ -19,6 +21,9 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.ResponseStatus;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Controller to expose methods for OPS personnel
  */
@@ -31,16 +36,19 @@ public class OpsController extends BaseController {
     private SubscriptionService subscriptionService;
     private CdrFileService cdrFileService;
     private MctsWsImportService mctsWsImportService;
+    private FrontLineWorkerService frontLineWorkerService;
     private EventRelay eventRelay;
 
 
     @Autowired
     public OpsController(SubscriptionDataService subscriptionDataService, SubscriptionService subscriptionService,
-                         CdrFileService cdrFileService, MctsWsImportService mctsWsImportService, EventRelay eventRelay) {
+                         CdrFileService cdrFileService, MctsWsImportService mctsWsImportService,
+                         FrontLineWorkerService frontLineWorkerService, EventRelay eventRelay) {
         this.subscriptionDataService = subscriptionDataService;
         this.subscriptionService = subscriptionService;
         this.cdrFileService = cdrFileService;
         this.mctsWsImportService = mctsWsImportService;
+        this.frontLineWorkerService = frontLineWorkerService;
         this.eventRelay = eventRelay;
     }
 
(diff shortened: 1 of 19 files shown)


koshalt:
@VishnuVimal @atish160384 This will be a good code review to watch for the NIC data consumption

api/src/main/java/org/motechproject/nms/api/web/OpsController.java
```diff
@@ -80,4 +92,59 @@ public void startUpkeep() {
         LOGGER.info("/upkeep");
         eventRelay.sendEventMessage(new MotechEvent(KilkariConstants.SUBSCRIPTION_UPKEEP_SUBJECT));
     }
+
+    @RequestMapping(value = "/addFlw",
```
larubbio: nitpick, but since the method creates or updates, shouldn't the URL endpoint reflect that?

api/src/main/java/org/motechproject/nms/api/web/OpsController.java
```diff
@@ -80,4 +92,59 @@ public void startUpkeep() {
...
+            method = RequestMethod.POST,
+            headers = { "Content-type=application/json" })
+    @ResponseStatus(HttpStatus.OK)
+    @Transactional
+    public void createUpdateFlw(@RequestBody AddFlwRequest addFlwRequest) {
+        // TODO: obscure phone number?
```
larubbio: Yes we should obscure the phone number, but it doesn't look like it is being logged so I think this comment can be removed.

flw/src/main/java/org/motechproject/nms/flw/domain/FlwError.java
```diff
@@ -0,0 +1,62 @@
+package org.motechproject.nms.flw.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+/**
+ * Audit log object for tracking failed FLW updates
+ */
+@Entity(tableName = "nms_flw_errors")
```
larubbio: Why is this table unique to this ops endpoint?  Do we have similar logging for the web services import?  (will I find that later in the code review?)

koshalt: Unique for now. The goal is to eventually refactor mcts sync to go through here and create the sync error objects.... but did not want to add that into this mix

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -150,6 +163,69 @@ public State getState(FrontLineWorker frontLineWorker) {
...
+
+        FrontLineWorker existingFlwByNumber = getByContactNumber(contactNumber);
+        FrontLineWorker existingFlwByMctsFlwId = getByMctsFlwIdAndState(mctsFlwId, state);
+        Map<String, Object> location = new HashMap<>();
+        try {
+            location = locationService.getLocations(flw);
+
+            if (existingFlwByMctsFlwId != null && existingFlwByNumber != null) {
+
+                if (existingFlwByMctsFlwId.getMctsFlwId().equalsIgnoreCase(existingFlwByNumber.getMctsFlwId()) &&
```
larubbio: This if check feels weird to me.  Basically there are two lookups to the database, one by mctsFlwId and the other by phone.  We then compare the results to verify the ids match and the two records are in the same state.  Is it possible to get two records back or can we just compare the primary key id?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -150,6 +163,69 @@ public State getState(FrontLineWorker frontLineWorker) {
...
+                        existingFlwByMctsFlwId.getState().equals(existingFlwByNumber.getState())) {
+                    // we are trying to update the same existing flw. set fields and update
+                    LOGGER.debug("Updating existing user with same phone number");
+                    update(FlwMapper.updateFlw(existingFlwByMctsFlwId, flw, location));
+                    return true;
+                } else {
```
larubbio: In this else block we assume the phone number is what is in conflict, but isn't it possible that the conflict is from mcts id as well?

Or what if I didn't have a phone number but I did have mcts flw id, or the reverse (i.e. I had a phone number but no id)  In both of those cases wouldn't we want to update the record and add the new information?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
```diff
@@ -150,6 +163,69 @@ public State getState(FrontLineWorker frontLineWorker) {
...
+                }
+            } else if (existingFlwByMctsFlwId != null && existingFlwByNumber == null) {
+                // trying to update the phone number of the person. possible migration scenario
+                // making design decision that flw will lose all progress when phone number is changed. Usage and tracking is not
+                // worth the effort & we don't really know that its the same flw
+                LOGGER.debug("Updating phone number for flw");
+                update(FlwMapper.updateFlw(existingFlwByMctsFlwId, flw, location));
+                return true;
+            } else if (existingFlwByMctsFlwId == null && existingFlwByNumber != null) {
+                // phone number used by someone else.
```
larubbio: Is it guarenteed to be the case that the phone number is in use by someone else?  Could we have just had the FLW without their mcts flw id?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
koshalt: That's good catch. Technically, I do need to catch all anonymous(existing) users without an mcts id and assign them the mcts update with the same phone number. **Need to fix**

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/MctsBeneficiaryUpdateServiceBundleIT.java
```diff
@@ -521,7 +521,7 @@ public void testUpdateBeneficiariesFromFile() throws Exception {
         assertEquals(453L, (long) updatedMother3.getHealthBlock().getCode());
 
         // DOB update:
-        String updatedDOB = "01-07-2015";
+        String updatedDOB = "01-12-2015";
```
larubbio: Is there a reason this date had to be changed?  Will we run into a time specific failure in the future or something?

larubbio:
A couple of small comments otherwise it looks ok.

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
koshalt: Resolved in person

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerServiceImpl.java
koshalt: Resolved in person. Added new clause and test

api/src/main/java/org/motechproject/nms/api/web/OpsController.java
koshalt: fixed

api/src/main/java/org/motechproject/nms/api/web/OpsController.java
koshalt: done. fixed.

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/MctsBeneficiaryUpdateServiceBundleIT.java
koshalt: yup, these keep constantly rolling off. we either have it in a file or manually write the string to a stream.

api/src/main/java/org/motechproject/nms/api/web/OpsController.java
```diff
@@ -80,4 +92,65 @@ public void startUpkeep() {
         LOGGER.info("/upkeep");
         eventRelay.sendEventMessage(new MotechEvent(KilkariConstants.SUBSCRIPTION_UPKEEP_SUBJECT));
     }
+
+    @RequestMapping(value = "/createUpdateFlw",
```
koshalt: @VishnuVimal @atish160384 please note this change in endpoint. should be minimal change (from /addFlw) to update in your client if you are using the latest build from here.

koshalt:
test this please

PR #692, opened 2016-07-12
Ops helper to look at scores in bookmarks

koshalt:
This should expose a get endpoint to fetch scores inside bookmark for a calling number


atish160384:
seems fine , merging branch to master.

PR #693, opened 2016-07-27
Mcts import retrial in the last 7 days

sripooja:
NMS-360
Introduce retrial of data import from MCTS service for 7 days

### DIFF
--- a/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
@@ -10,6 +10,7 @@ import org.motechproject.nms.csv.utils.GetLong;
 import org.motechproject.nms.csv.utils.GetString;
 import org.motechproject.nms.flw.domain.FrontLineWorker;
 
+import org.motechproject.nms.flw.exception.FlwExistingRecordException;
 import org.motechproject.nms.flw.exception.FlwImportException;
 import org.motechproject.nms.flw.service.FrontLineWorkerImportService;
 import org.motechproject.nms.flw.service.FrontLineWorkerService;
@@ -72,14 +73,14 @@ public class FrontLineWorkerImportServiceImpl implements FrontLineWorkerImportSe
             }
         } catch (ConstraintViolationException e) {
             throw new CsvImportDataException(createErrorMessage(e.getConstraintViolations(), csvImporter.getRowNumber()), e);
-        } catch (InvalidLocationException | FlwImportException | JDODataStoreException e) {
+        } catch (InvalidLocationException | FlwImportException | JDODataStoreException | FlwExistingRecordException e) {
             throw new CsvImportDataException(createErrorMessage(e.getMessage(), csvImporter.getRowNumber()), e);
         }
     }
 
     @Override
     @Transactional
-    public void importFrontLineWorker(Map<String, Object> record, State state) throws InvalidLocationException {
+    public void importFrontLineWorker(Map<String, Object> record, State state) throws InvalidLocationException, FlwExistingRecordException {
         FrontLineWorker flw = flwFromRecord(record, state);
 
         record.put(FlwConstants.STATE_ID, state.getCode());
@@ -92,6 +93,8 @@ public class FrontLineWorkerImportServiceImpl implements FrontLineWorkerImportSe
             //It updated_date_nic from mcts is not null,then it's not a new record. Compare it with the record from database and update
             if (mctsUpdatedDateNic != null && (flw.getUpdatedDateNic() == null || mctsUpdatedDateNic.isAfter(flw.getUpdatedDateNic()) || mctsUpdatedDateNic.isEqual(flw.getUpdatedDateNic()))) {
                 frontLineWorkerService.update(updateFlw(flw, record, location));
+            } else {
+                throw new FlwExistingRecordException("Updated record exists in the database");
             }
         }
     }
--- a/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
@@ -1,6 +1,7 @@
 package org.motechproject.nms.flw.service.impl;
 
 import org.apache.commons.lang.StringUtils;
+import org.joda.time.LocalDate;
 import org.motechproject.nms.csv.exception.CsvImportDataException;
 import org.motechproject.nms.csv.utils.ConstraintViolationUtils;
 import org.motechproject.nms.csv.utils.CsvImporterBuilder;
@@ -87,7 +88,11 @@ public class FrontLineWorkerImportServiceImpl implements FrontLineWorkerImportSe
         if (flw == null) {
             frontLineWorkerService.add(createFlw(record, location));
         } else {
-            frontLineWorkerService.update(updateFlw(flw, record, location));
+            LocalDate mctsUpdatedDateNic = (LocalDate) record.get(FlwConstants.UPDATED_ON);
+            //It updated_date_nic from mcts is not null,then it's not a new record. Compare it with the record from database and update
+            if (mctsUpdatedDateNic != null && (flw.getUpdatedDateNic() == null || mctsUpdatedDateNic.isAfter(flw.getUpdatedDateNic()) || mctsUpdatedDateNic.isEqual(flw.getUpdatedDateNic()))) {
+                frontLineWorkerService.update(updateFlw(flw, record, location));
+            }
         }
     }
 
--- a/flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
@@ -44,6 +44,12 @@ public class FrontLineWorkerImportController {
     @RequestMapping(value = "/update/language", method = RequestMethod.POST)
     @ResponseStatus(HttpStatus.OK)
     public void updateFrontLineWorkersLanguage(@RequestParam MultipartFile csvFile) {
+
+        // Update FrontLineWorkers Language from csv import is not supported
+        if(true) {
+            throw new UnsupportedOperationException("MCTS FLW Language csv import is not supported.");
+        }
+
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 flwUpdateImportService.importLanguageData(new InputStreamReader(in));
@@ -63,6 +69,12 @@ public class FrontLineWorkerImportController {
     @RequestMapping(value = "/update/msisdn", method = RequestMethod.POST)
     @ResponseStatus(HttpStatus.OK)
     public void updateFrontLineWorkersMSISDN(@RequestParam MultipartFile csvFile) {
+
+        // Update FrontLineWorkers by Msisdn import from csv is not supported
+        if(true) {
+            throw new UnsupportedOperationException("MCTS FLW csv import is not supported.");
+        }
+
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 flwUpdateImportService.importMSISDNData(new InputStreamReader(in));
@@ -82,6 +94,12 @@ public class FrontLineWorkerImportController {
     @RequestMapping(value = "/import", method = RequestMethod.POST)
     @ResponseStatus(HttpStatus.OK)
     public void importFrontLineWorkers(@RequestParam MultipartFile csvFile) {
+
+        // Import FrontLineWorkers from csv is not supported
+        if(true) {
+            throw new UnsupportedOperationException("MCTS FLW csv import is not supported.");
+        }
+
         try {
             try (InputStream in = csvFile.getInputStream()) {
                 frontLineWorkerImportService.importData(new InputStreamReader(in));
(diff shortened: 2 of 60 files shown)


motech-gerrit:

koshalt:
ok to test

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
```diff
@@ -87,7 +88,11 @@ public void importFrontLineWorker(Map<String, Object> record, State state) throw
         if (flw == null) {
             frontLineWorkerService.add(createFlw(record, location));
         } else {
-            frontLineWorkerService.update(updateFlw(flw, record, location));
+            LocalDate mctsUpdatedDateNic = (LocalDate) record.get(FlwConstants.UPDATED_ON);
+            //It updated_date_nic from mcts is not null,then it's not a new record. Compare it with the record from database and update
+            if (mctsUpdatedDateNic != null && (flw.getUpdatedDateNic() == null || mctsUpdatedDateNic.isAfter(flw.getUpdatedDateNic()) || mctsUpdatedDateNic.isEqual(flw.getUpdatedDateNic()))) {
+                frontLineWorkerService.update(updateFlw(flw, record, location));
```
koshalt: you would need some sort of audit here to reflect that the update will be dropped in the else case, right?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -160,6 +161,12 @@ public boolean importMotherRecord(Map<String, Object> record) {
         Boolean abortion = (Boolean) record.get(KilkariConstants.ABORTION);
         Boolean stillBirth = (Boolean) record.get(KilkariConstants.STILLBIRTH);
         Boolean death = (Boolean) record.get(KilkariConstants.DEATH);
+        LocalDate mctsUpdatedDateNic = (LocalDate) record.get(KilkariConstants.LAST_UPDATE_DATE);
+
+        //validate if it's an updated record compared to one from database
+        if (mctsUpdatedDateNic == null || (mother.getUpdatedDateNic() != null && mother.getUpdatedDateNic().isAfter(mctsUpdatedDateNic))) {
+            return true;
```
koshalt: why is this short circuiting and returning true here?  Don't you want to actually update the mother record if this is true?

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
```diff
@@ -106,24 +116,36 @@ public void importFromMcts(List<Long> stateIds, LocalDate referenceDate, URL end
...
 
         LOGGER.info("Initiated import workflow from MCTS for mothers, children and ashas");
     }
 
+    private void sendImportEventForAUserType(Long stateId, MctsUserType userType, LocalDate referenceDate, URL endpoint, String importSubject) {
+
+        LOGGER.debug("Fetching all the failed imports in the last 7days for stateId {} and UserType {}", stateId, userType);
+        QueryParams queryParams = new QueryParams(new Order("importDate", Order.Direction.ASC));
+        List<MctsImportFailRecord> failedImports = mctsImportFailRecordDataService.getByStateAndImportdateAndUsertype(stateId, referenceDate.minusDays(6), userType, queryParams);
+        LocalDate startDate = failedImports.isEmpty() ? referenceDate : failedImports.get(0).getImportDate();
```
koshalt: Are you sure that the previous query always returns back an ordered list? get(0) looks dangerous

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
```diff
@@ -375,7 +414,36 @@ private MctsImportAudit saveImportedAnmAshaData(AnmAshaDataSet anmAshaDataSet, S
...
+    }
+
+    private void deleteMctsImportFailRecords(final LocalDate startReferenceDate, final LocalDate endReferenceDate, final MctsUserType mctsUserType, final Long stateId) {
+
+        @SuppressWarnings("unchecked")
+        SqlQueryExecution<Long> queryExecution = new SqlQueryExecution<Long>() {
+
+            @Override
+            public String getSqlQuery() {
+                String query = "DELETE FROM nms_mcts_failures WHERE importDate between :startReferenceDate AND :endReferenceDate AND stateCode = :stateId AND userType = :mctsUserType";
```
koshalt: I would not recommend running straight sql queries unless you have complex joins that force you to do it. Pull this data through MDS queries is pretty simple and you can iterate over the result set and delete them.

Also, you will have to broadcast an event to flush caches across all instance if you are bypassing MDS with a query like above.

mcts/src/main/resources/MctsAuditModification.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE motech_data_services.nms_mcts_audit CHANGE importDate endImportDate DATE NULL;
```
koshalt: This really isn't safe. What is the reason to change columns names? You will impact all reports, potentially ETL (though we don't have any) and this will force to you take downtime across all three instances.

Also, this break back compat with previous data models and is not good service design.

mcts/src/test/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImplTest.java
```diff
@@ -83,6 +84,7 @@
     private final LocalDate today = DateUtil.today();
     private final LocalDate yesterday = today.minusDays(1);
 
+    @Ignore
```
koshalt: Is this temporarily ignored?

mcts/src/main/java/org/motechproject/nms/mcts/domain/MctsImportAudit.java
```diff
@@ -11,7 +11,10 @@
 public class MctsImportAudit {
 
     @Field
-    private LocalDate importDate;
+    private LocalDate startImportDate;
```
koshalt: pretty dangerous to change schema names in a live production system. This also breaks back compat. I wouldn't recommend this unless there is no other way around it.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
sripooja: We are short circuiting in the cases where an update is not required. Cases like if it imports in a date range and if response has both updated and old record, then old record should not override updated record

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
sripooja: In order to make it return ordered list, we send a queryparam

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
sripooja: Sure. I will make the change

mcts/src/main/resources/MctsAuditModification.sql
sripooja: Instead of just importDate, from now on we will have startImportDate and endImportDate. So planned to modify the name. 
Will look for an alternative now

mcts/src/test/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImplTest.java
sripooja: Yes. We are working on IT cases now

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
sripooja: I will add a logger here. Else case is for records with older data. If updated ones are imported we ignore the older records from the same file. 
This case can occur if there are any failed imports in the past 7 days

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -183,6 +185,11 @@ public boolean importMotherRecord(Map<String, Object> record) {
             return false;
         }
 
+        //validate if it's an updated record compared to one from database
+        if (mother.getUpdatedDateNic() != null && (mctsUpdatedDateNic == null || mother.getUpdatedDateNic().isAfter(mctsUpdatedDateNic))) {
+            return true;
```
koshalt: dont you want to return false here because you are dropping the update? Higher up the call stack, this true or false ends up changing the count of records that was accepted/rejected for an mcts pull, right? this behavior is similar to the 3 if checks above right?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -243,6 +251,11 @@ public boolean importChildRecord(Map<String, Object> record) {
             return false;
         }
 
+        //validate if it's an updated record compared to one from database
+        if (child.getUpdatedDateNic() != null && (mctsUpdatedDateNic == null || child.getUpdatedDateNic().isAfter(mctsUpdatedDateNic))) {
+            return true;
+        }
```
koshalt: same as the mother comment. if the update is being dropped, don't you want to return false here?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
sripooja: I drop an update whenever its an older record and we have an updated record in the database.
This is the case when we import within a range of importDates. And why I increment the accepted count in this case is
1)Ideally if the previous import is successful, accepted count would have that older record. 
2)Rejected count has all the hpdviolation check and errors if any in saving the importData. But these older records doesn't fall in this category

flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
```diff
@@ -63,6 +64,12 @@ public void updateFrontLineWorkersLanguage(@RequestParam MultipartFile csvFile)
     @RequestMapping(value = "/update/msisdn", method = RequestMethod.POST)
     @ResponseStatus(HttpStatus.OK)
     public void updateFrontLineWorkersMSISDN(@RequestParam MultipartFile csvFile) {
+
+        // Update FrontLineWorkers by Msisdn import from csv is not supported
+        if(true) {
```
koshalt: No need for the if(true) here. Just throw the exception and remove the rest of the code.

flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
```diff
@@ -82,6 +89,12 @@ public void updateFrontLineWorkersMSISDN(@RequestParam MultipartFile csvFile) {
     @RequestMapping(value = "/import", method = RequestMethod.POST)
     @ResponseStatus(HttpStatus.OK)
     public void importFrontLineWorkers(@RequestParam MultipartFile csvFile) {
+
+        // Import FrontLineWorkers from csv is not supported
+        if(true) {
```
koshalt: same as the above. just throw the exception and remove the rest of the code.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
koshalt: Your #2 assumption is invalid. The rejected count has everything in a feed that didn't pass business rules to make it to the db. In this case, your update feed didn't make it into the db because it failed a new business logic (updated_date) that you have introduced. If it was up to me, I would merge this with the rejected count for the feed or introduce a new paradigm to track duplicate old updates. 

Returning true here breaks all existing logic and assumptions around counts.

kilkari/src/main/java/org/motechproject/nms/kilkari/web/MctsBeneficiaryImportController.java
```diff
@@ -63,6 +63,11 @@ public void setCsvAuditService(CsvAuditService csvAuditService) {
     @ResponseStatus(HttpStatus.OK)
     public void importMotherData(@RequestParam MultipartFile csvFile) {
 
+        // Mcts mother import from csv is not supported
+        if(true) {
```
koshalt: Not a fan of this coding style. If you are really trying to remove this, you should do it the right way instead of an always true condition and leaving coding artifacts around. This feels like an easy-lazy solution that is not good design.

mcts/src/main/resources/MctsAuditModification.sql
koshalt: do you still need this file to be checked in?

koshalt:
A couple of minor things that you probably want to clean up. looks ok to merge otherwise.

flw/src/main/java/org/motechproject/nms/flw/web/FrontLineWorkerImportController.java
atish160384: I want to merge this thing later in my spare time, which can be deployed along with next deployment.That's why we decided not to remove this thing.That won't  affect our regular deadlines.

koshalt:
looks good. ok to merge and deploy

PR #694, opened 2016-08-22
Fixing the bug, when ACTIVE child is present, mother should stay deactivated

VishnuVimal:
…

### DIFF
--- a/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
+++ b/flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
@@ -8,6 +8,7 @@ import org.motechproject.nms.csv.utils.CsvImporterBuilder;
 import org.motechproject.nms.csv.utils.CsvMapImporter;
 import org.motechproject.nms.csv.utils.GetLong;
 import org.motechproject.nms.csv.utils.GetString;
+import org.motechproject.nms.csv.utils.GetLocalDate;
 import org.motechproject.nms.flw.domain.FrontLineWorker;
 
 import org.motechproject.nms.flw.exception.FlwExistingRecordException;
@@ -182,6 +183,7 @@ public class FrontLineWorkerImportServiceImpl implements FrontLineWorkerImportSe
         mapping.put(FlwConstants.VILLAGE_NAME, new Optional(new GetString()));
 
         mapping.put(FlwConstants.TYPE, new Optional(new GetString()));
+        mapping.put(FlwConstants.UPDATED_ON, new Optional(new GetLocalDate()));
 
         return mapping;
     }
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
@@ -19,6 +19,7 @@ import org.motechproject.nms.kilkari.domain.SubscriptionPack;
 import org.motechproject.nms.kilkari.domain.SubscriptionPackType;
 import org.motechproject.nms.kilkari.domain.SubscriptionRejectionReason;
 import org.motechproject.nms.kilkari.domain.SubscriptionStatus;
+import org.motechproject.nms.kilkari.exception.MultipleSubscriberException;
 import org.motechproject.nms.kilkari.repository.MctsMotherDataService;
 import org.motechproject.nms.kilkari.repository.SubscriptionErrorDataService;
 import org.motechproject.nms.kilkari.service.MctsBeneficiaryImportService;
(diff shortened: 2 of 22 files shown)


motech-gerrit:

koshalt:
ok to test

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
```diff
@@ -182,6 +183,7 @@ private String readLineWhileNotBlank(BufferedReader bufferedReader) throws IOExc
         mapping.put(FlwConstants.VILLAGE_NAME, new Optional(new GetString()));
 
         mapping.put(FlwConstants.TYPE, new Optional(new GetString()));
+        mapping.put(FlwConstants.UPDATED_ON, new Optional(new GetLocalDate()));
```
koshalt: what is this used for?

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -190,6 +195,11 @@ public boolean importMotherRecord(Map<String, Object> record) {
             return false;
         }
 
+        //validate if an ACTIVE child is already present for the mother. If yes, ignore the update
+        if (childAlreadyPresent(mother.getBeneficiaryId())) {
+            return false;
```
koshalt: I would think that you would want to log to subscription error here, right? That way, in the future (after child is done) when bbc comes back and asks why a mother wasn't created.. you have an audit trail.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -425,4 +425,32 @@ public void deactivateAllSubscriptionsForSubscriber(long callingNumber) {
...
+
+                if (fqr.isEmpty()) {
+                    return null;
+                }
+
+                if (fqr.size() == 1) {
+                    return (Subscriber) fqr.get(0);
+                }
+
+                throw new IllegalStateException(String.format(KilkariConstants.MORE_THAN_ONE_SUBSCRIBER, motherId));
```
koshalt: this is a pretty dangerous design pattern. If your data ever ends up in this state in your db due to manual updates, this will blow up and never recover. You will fail the rest of the sync and potentially even more since the original call of this method is marked as transactional. 

I understand that you are following patterns established in this file by other people, but you should probably have a catch on this particular exception and log an alert for this particular record. I dont think you want to fail the whole sync because of this. 

Also, this will be compounded by exception retries (x3) by activemq.

testing/src/test/java/org/motechproject/nms/testing/it/kilkari/MctsBeneficiaryImportServiceBundleIT.java
```diff
@@ -1420,4 +1420,45 @@ public void testImportIntoStateWithTwoCircles() throws Exception {
...
+        Subscriber subscriber = subscriberDataService.findByNumber(9439986187L);
+        Set<Subscription> subscriptions = subscriber.getAllSubscriptions();
+
+        Subscription childSubscription = subscriptionService
+                .getActiveSubscription(subscriber, SubscriptionPackType.CHILD);
+        Subscription pregnancySubscription = subscriptionService
+                .getActiveSubscription(subscriber, SubscriptionPackType.PREGNANCY);
+
+        //the mother subscription should still be DEACTIVATED as child is ACTIVE
+        assertEquals(2, subscriptions.size());
```
koshalt: Maybe I'm confused, but why would there be 2 subscriptions on the number? Mother first, then child gets activated along with the disabling with the mother. Then you import the mother again and wouldn't create a subscription for it. 

At the end of it, you will only have 1 child subscription on the number, right? Maybe you also want to check that out of the 2 subscriptions, only the child is active?

koshalt:
@VishnuVimal there are a few comments that you might want to address before merge. Also, this PR also seems to have partial artifacts from mcts retry/RCH? why is updated_on being added here?

flw/src/main/java/org/motechproject/nms/flw/service/impl/FrontLineWorkerImportServiceImpl.java
sripooja: Csv Import which are blocked from mcts retrial are fixed here.So added Updated_On field for mapping and respective testcases.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -95,7 +96,7 @@ public Subscriber execute(Query query) {
                 if (fqr.size() == 1) {
                     return (Subscriber) fqr.get(0);
                 }
-                throw new IllegalStateException(String.format(KilkariConstants.MORE_THAN_ONE_SUBSCRIBER, callingNumber));
+                throw new IllegalStateException(String.format(KilkariConstants.MORE_THAN_ONE_SUBSCRIBER_WITH_SAME_NUMBER, callingNumber));
```
koshalt: I want to make sure this exception being thrown is handled in our mcts import code as well. Might be unrelated to this feature PR though.

koshalt:
LGTM

PR #696, opened 2016-09-16
Nms 368

sripooja:
Code re-design for MSISDN update of Asha users in MA course


motech-gerrit:

sripooja:
@koshalt  These changes doesn't include testcases.

koshalt:
ok to test

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/MtrainingModuleActivityRecordAudit.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.jdo.annotations.Column;
+import javax.jdo.annotations.Unique;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+
+/**
+ * Audit table to track changes in MTRAINING_MODULE_ACTIVITYRECORD
+ */
+@Entity(tableName = "MTRAINING_MODULE_ACTIVITYRECORD_Audit")
+public class MtrainingModuleActivityRecordAudit extends MdsEntity {
```
koshalt: Is there a reason why this is a separate entity? Why not just use the track changes annotation to track this, like other places in code?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -457,4 +470,60 @@ public String getScoresForUser(Long callingNumber) {
...
+        if (existingBookmarks.size() > 0) {
+            int i;
+            Bookmark bookmark;
+            for (i = 0; i < existingBookmarks.size(); i++) {
+                bookmark = existingBookmarks.get(i);
+                bookmark.setExternalId(newCallingNumber.toString());
+                bookmarkService.updateBookmark(bookmark);
+            }
+            LOGGER.debug("Updated MSISDN {} to {} in {} Bookmarks", oldCallingNumber, newCallingNumber, i);
+        } else {
```
koshalt: For here and below, how about checks first to see if the new phone number has existing bookmarks and completion records already? What to do then?

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
```diff
@@ -457,4 +470,60 @@ public String getScoresForUser(Long callingNumber) {
...
+        List<ActivityRecord> existingRecords = activityDataService.findRecordsForUser(oldCallingNumber.toString());
+        if (existingRecords.size() > 0) {
+            int i;
+            ActivityRecord activityRecord;
+            for (i = 0; i < existingRecords.size(); i++) {
+                activityRecord = existingRecords.get(i);
+                activityRecord.setExternalId(newCallingNumber.toString());
+                activityDataService.update(activityRecord);
+            }
+            mtrainingModuleActivityRecordAuditDataService.create(new MtrainingModuleActivityRecordAudit(oldCallingNumber, newCallingNumber));
```
koshalt: Why is activity record the only one entity being tracked with an audit? What about  completion record updates? Again, probably want to try and use the track changes annotation instead of creating completely new data services and entities.

flwUpdate/src/main/java/org/motechproject/nms/flwUpdate/service/impl/FrontLineWorkerImportServiceImpl.java
```diff
@@ -94,12 +105,91 @@ public void importFrontLineWorker(Map<String, Object> record, State state) throw
             //It updated_date_nic from mcts is not null,then it's not a new record. Compare it with the record from database and update
             if (mctsUpdatedDateNic != null && (flw.getUpdatedDateNic() == null || mctsUpdatedDateNic.isAfter(flw.getUpdatedDateNic()) || mctsUpdatedDateNic.isEqual(flw.getUpdatedDateNic()))) {
                 frontLineWorkerService.update(updateFlw(flw, record, location));
+                Long oldMsisdn = flw.getContactNumber();
+                Long newMsisdn = (Long) record.get(FlwConstants.CONTACT_NO);
```
sripooja: Will add a check there. We don't have a confirmation but it's a mandatory field in mcts and the same is observed in RCH responses as well.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
sripooja: Since we are working only on the MSISDN change related issues, we added to ActivityRecord. Issues related to Completion record will be fixed in the corresponding tickets.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/domain/MtrainingModuleActivityRecordAudit.java
sripooja: we can't add annotation to track ActivityRecord as domain belongs to motech.

testing/src/test/java/org/motechproject/nms/testing/it/flwUpdate/FrontLineWorkerUpdateImportServiceBundleIT.java
```diff
@@ -723,4 +744,40 @@ public void verifyFT561() throws InterruptedException, IOException {
...
+        bookmarkDataService.create(new Bookmark(contactNumber.toString(), "1", "1", "1", new HashMap<String, Object>()));
+
+        CompletionRecord cr = new CompletionRecord(contactNumber, 35, false, 1);
+        completionRecordDataService.create(cr);
+//        String externalId, String courseName, String chapterName, String lessonName, DateTime startTime, DateTime completionTime, ActivityState.STARTED);
+        ActivityRecord ar = new ActivityRecord(contactNumber.toString(), "1", "1", "1", null,null , ActivityState.STARTED);
+        activityDataService.create(ar);
+    }
+
+    private void assertMaRecords(Long oldContactNumber, Long newContactNumber) {
```
koshalt: Will there be tests for other edge cases and negative test cases? Looks like we have exactly 1 test to traverse  the positive code path here.

mobile-academy/src/main/java/org/motechproject/nms/mobileacademy/service/impl/MobileAcademyServiceImpl.java
sripooja: We can have the ideal case once the database is fixed (NMS-369). But we can't wait till the issue is fixed as Rajasthan State Govt is going to release BSNL SIM cards for all the FLWs in the state. Thus we need to deploy the code changes immediately. We do not expect these numbers to overlap with existing MSISDNs, thus BBCMA has been insisting that we deploy the code changes ASAP.

testing/src/test/java/org/motechproject/nms/testing/it/flwUpdate/FrontLineWorkerUpdateImportServiceBundleIT.java
sripooja: Added one negative case since for all negative code path irrespective of Bookmark/Completion/ActivityRecord Update, Flw update will be rejected.

koshalt:
Ok to go to staging and then production. Don't forget to create a release build for it as well. Thanks.

PR #698, opened 2016-10-10
Asha validation for FLWs (NMS-372)

sripooja:
This pull request is to add an extra check in the code so that it does not allow any FLW record with Type other than ASHA.


motech-gerrit:

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -46,6 +46,7 @@
     public static final String NOT_FOUND = "<%s: Not Found>";
     public static final String NOT_AUTHORIZED = "<%s: Not Authorized>";
     public static final String NOT_DEPLOYED = "<%s: Not Deployed In State>";
+    public static final String INVALID_TYPE = "<%s: Invalid Type>";
```
koshalt: Was this agreed to with IMI? Why introduce this as opposed to  not authorized? IMI tends to have a direct dependency on these return types.

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -59,6 +60,7 @@
     public static final int MA_MIN_SCORE = 0;
     public static final int MA_MAX_SCORE = 4;
     public static final String CALLING_NUMBER = "callingNumber";
+    public static final String VALID_TYPE = "ASHA";
```
koshalt: is this guaranteed to be upper case all the time?

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
```diff
@@ -392,24 +392,30 @@ private MctsImportAudit saveImportedAnmAshaData(AnmAshaDataSet anmAshaDataSet, S
...
-                rejected++;
-            } catch (FlwExistingRecordException e) {
-                LOGGER.error("Cannot import FLW with ID: {}, and MSISDN (Contact_No): {}", record.getId(), record.getContactNo(), e);
-                rejected++;
-            } catch (Exception e) {
-                LOGGER.error("Flw import Error. Cannot import FLW with ID: {}, and MSISDN (Contact_No): {}",
-                        record.getId(), record.getContactNo(), e);
+            String designation = record.getType();
+            designation = (designation != null) ? designation.trim() : designation;
+            if (!("ASHA".equals(designation))) {
```
koshalt: introduce the string literal as a constant somewhere and reuse it.

testing/src/test/java/org/motechproject/nms/testing/it/api/OpsControllerBundleIT.java
```diff
@@ -170,11 +170,19 @@ public void testBadContactNumberAddFlwRequest() throws IOException, InterruptedE
...
     @Test
     public void testCreateNewFlw() throws IOException, InterruptedException {
-        AddFlwRequest addFlwRequest = getAddRequest();
+        AddFlwRequest addFlwRequest = getAddRequestASHA();
         HttpPost httpRequest = RequestBuilder.createPostRequest(addFlwEndpoint, addFlwRequest);
         assertTrue(SimpleHttpClient.execHttpRequest(httpRequest, HttpStatus.SC_OK, RequestBuilder.ADMIN_USERNAME, RequestBuilder.ADMIN_PASSWORD));
     }
 
+    // Validating whether the type is ASHA or not
+    @Test
```
koshalt: Is there a negative test to confirm that this fails if the designation is wrong?

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
sripooja: These are not return types. They are just to print error for debugging purposes internally. So, we don't need to reach out to IMI.

VishnuVimal: These are not return types. They are just to print error for debugging purposes internally. So, we don't need to reach out to IMI.

api/src/main/java/org/motechproject/nms/api/web/BaseController.java
```diff
@@ -124,6 +128,18 @@ protected static boolean validateField10Digits(StringBuilder errors, String fiel
         return false;
     }
 
+    protected static boolean validatetypeASHA(StringBuilder errors, String fieldName, String mctsFlwId, Long contactNumber, String type) {
```
koshalt: I think I understand where the confusion is now. The bug primarily only talks about filtering asha users for import. And this validation exists in the controller, which is used only for incoming/outgoing calls to IMI.

If we are only dealing with imports, this is not the right place for the validation.
If we are only dealing with live call filtering, then this is ok and you would add these associated check to the UserController

flw/src/main/java/org/motechproject/nms/flw/utils/FlwConstants.java
```diff
@@ -22,6 +22,7 @@
     public static final String NON_CENSUS_VILLAGE_ID = "SVID";
     public static final String TYPE = "Type";
     public static final String UPDATED_ON = "Updated_On";
+    public static final String VALID_TYPE = "ASHA";
```
koshalt: non-blocking comment - but this is not potentially always true right? Valid type could be extended to asha & dosha tomorrow. and then you will have to check for designation.equals asha || dosha, etc. 

Just call it ASHA_TYPE in the future.

PR #699, opened 2016-11-06
Csv import fix

sripooja:
NMS-374 Refactor Kilkari CSV Import so that the whole import is not halted because of an error in a single record


motech-gerrit:

beehyvajai:
Looks good.

PR #701, opened 2016-12-15
Added Rch import controller, service and modified respective testcases

sripooja:
Implement RCH CSV import


motech-gerrit:

PR #702, opened 2016-12-16
NMS-375 Stop purging of deactivated subscritions

anuranjan:
Removed the purging method call.


motech-gerrit:

PR #703, opened 2016-12-22
Fixed Msisdn Update in MA tables

sripooja:
NMS-368 Code re-design for MSISDN update of Asha users in MA course


motech-gerrit:

PR #718, opened 2018-05-22
corrections in child import, removing nullpointer exceptions

VishnuVimal:


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -364,7 +364,6 @@ public ChildImportRejection importChildRecord(Map<String, Object> record, Subscr
             action = (String) record.get(KilkariConstants.ACTION);
             childId = (String) record.get(KilkariConstants.RCH_ID);
             child = (MctsChild) record.get(KilkariConstants.RCH_CHILD);
-            LOGGER.debug("Child:{}", child.getName());
```
gudipatiharitha: Is it possible for child to be null -- add null check for child

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -640,7 +640,11 @@ public Subscription reactivateSubscription(Subscriber subscriber, Subscription s
         subscription.setStatus(SubscriptionStatus.ACTIVE);
         subscription.setDeactivationReason(null);
         subscriptionService.updateStartDate(subscription, dateTime);
-        subscriptionService.deleteBlockedMsisdn(subscriber.getMother().getId(), null, subscriber.getCallingNumber());
+        if (subscriber.getMother() != null) {
```
gudipatiharitha: add nnull check above previous line, since we refer getId from mother in that line too

PR #720, opened 2018-06-04
child update with mother null bug fix

VishnuVimal:
I've fixed the bug where child with mother null is overriding previous mother, if present

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
@@ -585,6 +585,7 @@ public class SubscriberServiceImpl implements SubscriberService {
                 if (subscribersByMsisdn.size() == 1 && childUpdate.getMother() != null && subscribersByMsisdn.get(0).getMother() != null && subscribersByMsisdn.get(0).getChild() == null) {
                     //update subscriber with child
                     if (subscribersByMsisdn.get(0).getMother().equals(childUpdate.getMother())) {
+                        childUpdate.setDateOfBirth(dob);
                         Subscriber subscriber = subscribersByMsisdn.get(0);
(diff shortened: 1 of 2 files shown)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -581,6 +585,7 @@ public ChildImportRejection updateRchChildSubscriber(Long msisdn, MctsChild chil
                 if (subscribersByMsisdn.size() == 1 && childUpdate.getMother() != null && subscribersByMsisdn.get(0).getMother() != null && subscribersByMsisdn.get(0).getChild() == null) {
                     //update subscriber with child
                     if (subscribersByMsisdn.get(0).getMother().equals(childUpdate.getMother())) {
+                        childUpdate.setDateOfBirth(dob);
```
gudipatiharitha: We should not be updating child DOB here, can you please check why and atleast add a TODO if other tasks have priority

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriberServiceImpl.java
```diff
@@ -475,6 +475,8 @@ public ChildImportRejection updateChildSubscriber(Long msisdn, MctsChild childUp
                         subscriberByMctsId.setDateOfBirth(dob);
                         if (subscriberByMctsId.getMother() == null) {
                             subscriberByMctsId.setMother(childUpdate.getMother());
+                        } else  if (subscriberByMctsId.getMother() != null && childUpdate.getMother() == null) {
+                            childUpdate.setMother(subscriberByMctsId.getMother());
```
gudipatiharitha: Looks like child need not be updated if mother is coming as null -- since childupdate.getmother came as null - we are updating again with old db value. Just check if childUpdate.setMother --doesn't set at all to null -- before calling this method

PR #723, opened 2018-07-04
feature_nic_new_locations

Sree21:
Added logger statements and minor performance improvements


rch/src/main/java/org/motechproject/nms/rch/service/impl/RchWebServiceFacadeImpl.java
```diff
@@ -1439,7 +1443,9 @@ private void childLocUpdateFromCsv(List<Map<String, Object>> result, Long stateI
                     locArrList.add(locMap);
                 }
             }
-            updateLocInMap(locArrList, stateId, rchUserType);
+            if (!locArrList.isEmpty()) {
+                updateLocInMap(locArrList, stateId, rchUserType);
+            }
 
         } catch (NullPointerException e) {
```
gudipatiharitha: Do not catch it -- add null check

rch/src/main/java/org/motechproject/nms/rch/service/impl/RchWebServiceFacadeImpl.java
```diff
@@ -1511,7 +1519,9 @@ private void ashaLocUpdateFromCsv(List<Map<String, Object>> result, Long stateId
...
```
gudipatiharitha: Do not catch it

PR #741, opened 2018-07-31
Locations web services and csv

Sree21:
Added code to import locations data from web services.
Added endpoint to read data from location CSVs.
Added code to bulk insert location records from XML or CSV to database.


rch/src/main/resources/rch.properties
```diff
@@ -10,7 +10,7 @@ rch.sync.cron=0 2 18 * * ? *
 rch.mother.sync.cron=0 0 20 * * ? *
 rch.child.sync.cron=0 10 20 * * ? *
 rch.asha.sync.cron=0 20 20 * * ? *
-rch.location.sync.cron=0 21 20 * * ? *
+rch.location.sync.cron=0 30 19 * * ? *
```
gudipatiharitha: cant we call mothers after location is done ?

PR #746, opened 2018-08-02
CSR Processing Performance

sudheer93:
increased csr processing performance by adding bulk insert and added unique requestID


imi/src/main/java/org/motechproject/nms/imi/service/impl/CdrFileServiceImpl.java
```diff
@@ -713,10 +716,7 @@ public void processCsrs(File file, int lineCount) { //NOPMD NcssMethodCount
                 try {
 
                     CallSummaryRecord csr = CsrHelper.csvLineToCsr(line);
-                    if (callSummaryRecordDataService.countFindByRequestId(csr.getRequestId()) == 0) {
-                        callSummaryRecordDataService.create(csr);
-                        saveCount++;
-                    }
+                    callSummaryRecords.add(csr);
```
gudipatiharitha: Might be better to clear callSummaryRecords after batch count - but since testedandworkingfine on 8 GB ram too -- approving it

PR #748, opened 2018-08-03
parallelisation of web service import

VishnuVimal:


rch/src/main/java/org/motechproject/nms/rch/service/impl/RchWebServiceFacadeImpl.java
```diff
@@ -662,71 +662,82 @@ public void readVillageResponseFromFile(MotechEvent event) throws RchFileManipul
...
-                        rchImportFailRecordDataService.create(new RchImportFailRecord(endDate, RchUserType.MOTHER, stateId));
-                    } catch (NullPointerException e) {
-                        LOGGER.error("No files saved a : ", e);
+                    } else {
+                        String warning = String.format("No Mother data set received from RCH for %d stateId", stateId);
+                        LOGGER.warn(warning);
                     }
+                } catch (JAXBException e) {
+                    throw new RchInvalidResponseStructureException(String.format("Cannot deserialize RCH mother data from %s location.", stateId), e);
+                } catch (RchInvalidResponseStructureException e) {
```
gudipatiharitha: there should never be such space difference in the commits !! tough to review ,. Please do git diff before ccommit

PR #774, opened 2019-01-17
Remove ignored tests

LakshmiAmbikaV:
1. Removed IVR location csv and mobile kunji tests, changed IVR to MCTS_IMPORT and fixed the tests.
2. Fixed MCTS and RCH failed tests by changing mobile number, beneficiary ID and RCH ID
3. Fixed MCTS and subscription related tests and Ignored one test


testing/src/test/java/org/motechproject/nms/testing/it/IntegrationTests.java
```diff
@@ -77,10 +77,11 @@
     /**
      * Kilkari
      */
+    RchBeneficiaryImportServiceBundleIT.class,
```
gudipatiharitha: This commit looks totally unnecessary - always commt from IDE after checking the differences manually

PR #775, opened 2019-01-17
Rejection reason MOBILE_NUMBER_ALREADY_IN_USE changed to MOBILE_NUMBE…

VishnuVimal:
…R_ALREADY_SUBSCRIBED


kilkari/src/main/java/org/motechproject/nms/kilkari/domain/RejectionReasons.java
```diff
@@ -6,7 +6,6 @@
 public enum RejectionReasons {
     FLW_TYPE_NOT_ASHA,
     INVALID_LOCATION,
-    MOBILE_NUMBER_ALREADY_IN_USE,
```
gudipatiharitha: why do I just see removal o fMOBILE_NUMBER_ALREADY_IN_USE from the enum, but not the other reason added to this ?

gudipatiharitha: looks like new one is already present !!

PR #776, opened 2019-01-18
logs for file already present for state and date rch

VishnuVimal:


rch/src/main/java/org/motechproject/nms/rch/service/impl/RchImportFacilitatorServiceImpl.java
```diff
@@ -37,7 +37,7 @@ public void createImportFileAudit(RchImportFacilitator rchImportFacilitator) thr
         if (rchImportFacilitator.getFileName() == null) {
             throw new RchFileManipulationException("Invalid file name");
         } else if (rchImportFacilitator1.size() != 0) {
-            LOGGER.error("A record already present for the same state and today's date.");
+            LOGGER.debug("A record already present for the same state: {} and today's date: {}.", state, importDate);
```
gudipatiharitha: also need the type !!!

PR #785, opened 2019-02-14
subscriber update

VishnuVimal:


flw/src/main/java/org/motechproject/nms/flw/domain/FrontLineWorker.java
```diff
@@ -77,6 +77,7 @@
 
     @Field
     @Persistent(defaultFetchGroup = "true")
+    @TrackField
```
gudipatiharitha: Why is this added ? Is ths working ??

rch/src/main/java/org/motechproject/nms/rch/service/impl/RchWebServiceFacadeImpl.java
```diff
@@ -325,7 +325,7 @@ private void retryScpFromRemoteToLocal(String name, String remoteLocation, Local
             } else {
                 for (RchImportFacilitator rchImportFacilitatorType: rchImportFacilitatorTypes
                 ) {
```
gudipatiharitha: Is this tested ? and abov nnull check

PR #808, opened 2020-07-10
Data Integrity Error Fix

jangidrakesh:
two root cause of data integrity are fixed in this branch
1)Scenario:
    creating a rch mother record with rch id r1 and mcts id=null
    update the mother record with rch id r1 and mcts id m1
   
    Expected result=> should update the mcts id in mother record
    Actual Result=> Reject the mother record with data integrity error
2)Scenario:
    import rch mother with rch_id r1+"\n"
    import rch mother with rch id r1
Whenever we receive update (without '\n') for records whose rchId incude"\n" at the end, we are treating update as a new subscriber 
-------------------------------------------------------------------FIX------------------------------------------------------------
**Issue**: The existing code allow line break(\n,\t,\r, white space) in the give ids(see the table) 
Eg:Import record 1 with above special char then import update for same record but this time id doesn't contain special char, In this scenario system treating this id as new Id but both ids are same in the term of numerical digits.

**After fix behavior of the system:** This fix will remove new Line('\n'), white space(' '), tab('\t'), '\r' form the given ids (table) before processing to DB,
Eg: if we import record with id="12 24\t5\n\t"
then system will save the record with id=12245

IMPORT Type               ID in Imported Record

RCH Mother                  Registration_no
                                      MCTS_ID_No

RCH Child                     Registration_no
                                      MCTS_ID_No
                                      Mother_Registration_no
                                      MCTS_Mother_ID_No

MCTS Mother               ID_No

MCTS Child                  ID_No
                                     Mother_ID

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/ChildCsvThreadProcessor.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/ChildCsvThreadProcessor.java
@@ -1,10 +1,7 @@
 package org.motechproject.nms.kilkari.service;
 
 import org.motechproject.metrics.service.Timer;
-import org.motechproject.nms.kilkari.domain.MctsChild;
-import org.motechproject.nms.kilkari.domain.RejectionReasons;
-import org.motechproject.nms.kilkari.domain.SubscriptionOrigin;
-import org.motechproject.nms.kilkari.domain.ThreadProcessorObject;
+import org.motechproject.nms.kilkari.domain.*;
 import org.motechproject.nms.kilkari.utils.KilkariConstants;
 import org.motechproject.nms.region.domain.LocationFinder;
 import org.motechproject.nms.rejectionhandler.domain.ChildImportRejection;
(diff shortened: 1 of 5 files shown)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/ChildCsvThreadProcessor.java
```diff
@@ -54,21 +51,64 @@ public ThreadProcessorObject call() throws Exception { //NOPMD SignatureDeclareT
...
+                try{
+                    String newMctsMotherId=(String)record.get(mctsMotherIdForChild);
+                    newMctsMotherId=newMctsMotherId.replaceAll("[\\n\\t\\r ]","");
+                    record.replace(mctsMotherIdForChild,newMctsMotherId);
+                }
+                catch (Exception e){
+                    //LOGGER.debug("no mcts mother id for child");
+                }
+            }
+
```
anuranjan, line 111: there are four replace logic, which can be abstracted out in a single method and the same method can be reused.
Also, just make sure there is no scope of null pointer in (String)record.get(xyz).

kilkari/src/main/java/org/motechproject/nms/kilkari/service/MotherCsvThreadProcessor.java
```diff
@@ -63,13 +64,30 @@ public ThreadProcessorObject call() throws Exception { //NOPMD SignatureDeclareT
...
+            if(!mctsImport){
+                try {
+                    String newMctsId=(String)record.get(mctsIDForRchMother);
+                    newMctsId=newMctsId.replaceAll("[\\n\\t\\r ]","");
+                    record.replace(mctsIDForRchMother,newMctsId);
+                }
+                catch (Exception e){
+                    //LOGGER.debug("no mcts id for the rch mother");
+                }
+            }
```
anuranjan, line 89: Here as well, the common replace method can be used. Have this new method in a utility class and access everywhere.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -303,6 +303,12 @@ public ChildImportRejection importChildRecord(Map<String, Object> record, Subscr
 
                 try {
                     MctsMother motherInstance = (MctsMother) motherRecord;
+                    //removing special char from mcts child's mother_id
+                    String mctsMotherId=motherInstance.getBeneficiaryId();
+                    mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");
+                    motherInstance.setBeneficiaryId(mctsMotherId);
+                    record.replace(KilkariConstants.MOTHER_ID,motherInstance);
```
anuranjan, line 310: Same here

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
```diff
@@ -320,6 +320,11 @@ public MctsImportAudit saveImportedMothersData(MothersDataSet mothersDataSet, St
             beneficiaryId = (String) recordMap.get(KilkariConstants.BENEFICIARY_ID);
             msisdn = (Long) recordMap.get(KilkariConstants.MSISDN);
             DateTime lmp = (DateTime) recordMap.get(KilkariConstants.LMP);
+
+            // removing special char from beneficiary id
+            beneficiaryId=beneficiaryId.replaceAll("[\\n\\t\\r ]","");
+            recordMap.replace(KilkariConstants.BENEFICIARY_ID,beneficiaryId);
+
```
anuranjan, line 327: same here

mcts/src/main/java/org/motechproject/nms/mcts/service/impl/MctsWsImportServiceImpl.java
```diff
@@ -493,6 +498,11 @@ private MctsImportAudit saveImportedChildrenData(ChildrenDataSet childrenDataSet
             childId = (String) recordMap.get(KilkariConstants.BENEFICIARY_ID);
             msisdn = (Long) recordMap.get(KilkariConstants.MSISDN);
             DateTime dob = (DateTime) recordMap.get(KilkariConstants.DOB);
+
+            // removing special char from beneficiary id
+            childId=childId.replaceAll("[\\n\\t\\r ]","");
+            recordMap.replace(KilkariConstants.BENEFICIARY_ID,childId);
+
```
anuranjan, line 505: same here

kilkari/src/main/java/org/motechproject/nms/kilkari/service/MotherCsvThreadProcessor.java
```diff
@@ -63,13 +64,30 @@ public ThreadProcessorObject call() throws Exception { //NOPMD SignatureDeclareT
...
+            record.replace(id,newRchId);
+            //for rch mother's MCTS_ID_No
+            if(!mctsImport){
+                try {
+                    String newMctsId=(String)record.get(mctsIDForRchMother);
+                    newMctsId=newMctsId.replaceAll("[\\n\\t\\r ]","");
+                    record.replace(mctsIDForRchMother,newMctsId);
+                }
+                catch (Exception e){
+                    //LOGGER.debug("no mcts id for the rch mother");
```
anuranjan, line 87: Why is this logger commented out? Exception should never be eaten up.

PR #809, opened 2020-07-12
Data integrity error fix Updated

jangidrakesh:
two root cause of data integrity are fixed in this branch
1)Scenario:
creating a rch mother record with rch id r1 and mcts id=null
update the mother record with rch id r1 and mcts id m1

Expected result=> should update the mcts id in mother record
Actual Result=> Reject the mother record with data integrity error
2)Scenario:
import rch mother with rch_id r1+"\n"
import rch mother with rch id r1
Whenever we receive update (without '\n') for records whose rchId incude"\n" at the end, we are treating update as a new subscriber
-------------------------------------------------------------------FIX------------------------------------------------------------
Issue: The existing code allow line break(\n,\t,\r, white space) in the give ids(see the table)
Eg:Import record 1 with above special char then import update for same record but this time id doesn't contain special char, In this scenario system treating this id as new Id but both ids are same in the term of numerical digits.

After fix behavior of the system: This fix will remove new Line('\n'), white space(' '), tab('\t'), '\r' form the given ids (table) before processing to DB,
Eg: if we import record with id="12 24\t5\n\t"
then system will save the record with id=12245

IMPORT Type ID in Imported Record

RCH Mother 
{Registration_no, MCTS_ID_No}

RCH Child 
{ Registration_no, MCTS_ID_No, Mother_Registration_no, MCTS_Mother_ID_No }

MCTS Mother { ID_No }

MCTS Child { ID_No, Mother_ID}

### DIFF
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
@@ -303,6 +303,12 @@ public class MctsBeneficiaryImportServiceImpl implements MctsBeneficiaryImportSe
 
                 try {
                     MctsMother motherInstance = (MctsMother) motherRecord;
+                    //removing special char from mcts child's mother_id
+                    String mctsMotherId=motherInstance.getBeneficiaryId();
+                    mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");
+                    motherInstance.setBeneficiaryId(mctsMotherId);
+                    record.replace(KilkariConstants.MOTHER_ID,motherInstance);
+
                     mother = mctsBeneficiaryValueProcessor.getMotherInstanceByBeneficiaryId(motherInstance.getBeneficiaryId());
                 } catch (Exception e) {
                     mother = mctsBeneficiaryValueProcessor.getMotherInstanceByBeneficiaryId(motherRecord.toString());
--- a/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
+++ b/kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
@@ -304,10 +304,10 @@ public class MctsBeneficiaryImportServiceImpl implements MctsBeneficiaryImportSe
                 try {
                     MctsMother motherInstance = (MctsMother) motherRecord;
                     //removing special char from mcts child's mother_id
-                    String mctsMotherId=motherInstance.getBeneficiaryId();
-                    mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");
-                    motherInstance.setBeneficiaryId(mctsMotherId);
-                    record.replace(KilkariConstants.MOTHER_ID,motherInstance);
+//                    String mctsMotherId=motherInstance.getBeneficiaryId();
+//                    mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");
+//                    motherInstance.setBeneficiaryId(mctsMotherId);
+//                    record.replace(KilkariConstants.MOTHER_ID,motherInstance);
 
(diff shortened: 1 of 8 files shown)


kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -303,6 +303,12 @@ public ChildImportRejection importChildRecord(Map<String, Object> record, Subscr
 
                 try {
                     MctsMother motherInstance = (MctsMother) motherRecord;
+                    //changes made to remove special character form the mcts child's mother_id
+                    String mctsMotherId=motherInstance.getBeneficiaryId();
+                    mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");
+                    motherInstance.setBeneficiaryId(mctsMotherId);
+                    record.replace(KilkariConstants.MOTHER_ID,motherInstance);
+
```
anuranjan, line 311: This should also be using the common method.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -977,8 +983,20 @@ public LocalDate retrieve(String value) {
 
         return mapping;
     }
+    public void removeSpecialChar(ArrayList<String > listOfIds,Map<String, Object> record){
```
anuranjan, line 986: Please move this method to some util class, and not a service class

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
```diff
@@ -977,8 +983,20 @@ public LocalDate retrieve(String value) {
...
 
+        for(String id :listOfIds){
+            try{
+                String idValue=(String)record.get(id);
+                idValue=idValue.replaceAll("[\\n\\t\\r ]","");
+                record.replace(id,idValue);
+            }
+            catch (Exception e){
+                continue;
```
anuranjan, line 995: Exception should never be eaten up. Please either log this or handle this.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
jangidrakesh: > This should also be using the common method.

**to replace this line**
 String mctsMotherId=motherInstance.getBeneficiaryId();
 mctsMotherId=mctsMotherId.replaceAll("[\\n\\t\\r ]","");

**we have to add these much of line** 

String mctsMotherId=motherInstance.getBeneficiaryId();
Map<String, Object> motherRecordTemp= new HashMap<>();;
motherRecordTemp.put("ID_No",mctsMotherId);
ArrayList<String > listOfIds=new ArrayList<>();
listOfIds.add("ID_No");
removeSpecialChar(listOfIds,record);
mctsMotherId=(String)motherRecordTemp.get("ID_No");

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
jangidrakesh: > Which exception is expected here? A more concrete Exception should be caught, and should never be eaten up. Please either log this or handle this.

Some of Ids may have null values(Not all ids are required in csv/xml records),  
So hasmap will throw an exception in that case and code will stopped.

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/MctsBeneficiaryImportServiceImpl.java
jangidrakesh: > Please move this method to some util class, and not a service class. Also, create a string constant for the regular expression instead of hardcoding here in the logic.

there is a method which validate referance date that's why I thought  this is a validation on ids
So i put it here

PR #828, opened 2021-06-10
splitting_cdr_csr

jangidrakesh:
by this fix we can process cdr/csr multiple times in a day


jangidrakesh:
I keep the commented code so It will be easy to revert back to one cdr/csr per day

PR #844, opened 2024-04-10
Bug resolved: Import process issue with location names

Sheetal-VB:
Due to extra char in name(String) and location names, insert query in rejection table is failing. Issue resolved.


atishopera:
Merge it

PR #858, opened 2024-10-03
Two or more subscriber having same calling number and active, pending or hold subscriptions

Sheetal-VB:


Kantas2601:
Bad Request

PR #870, opened 2024-12-19
Target file performance Improvement.

dinesh-beehyv:
Below changes has been done to inprove the performance
Approach for Performance Improvement:
Fetching data using a JDBC connection instead of Data Nucleus.
Optimized Query Execution:
Using prepared statements for executing database queries.
Ensuring all operations are properly closed after a successful database interaction.
Query Optimization:
Updated the existing query to fetch all the required columns from 9 tables used for target file generation.
DTO Usage:
Storing query results in a DTO format.
Applied the Builder design pattern in the DTO class for efficient data handling.
Integration with Method:
Updating all columns in the generateFreshCall method.

### DIFF
--- a/imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
+++ b/imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
@@ -632,7 +632,7 @@ public class TargetFileServiceImpl implements TargetFileService {
 
 
 
-    List<SubscriptionDto> subscriptions = subscriptionService.findActiveSubscriptionsForDayApi(dow, offset, maxQueryBlock);
+    List<SubscriptionDto> subscriptions = findAllSubscriptionsForDayApi(dow, maxQueryBlock);
 
             LOGGER.info("Subs_block_size" + subscriptions.size());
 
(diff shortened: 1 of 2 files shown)


imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
```diff
@@ -577,50 +604,54 @@ private HashMap<String, Integer> generateFreshCalls(DateTime timestamp, int maxQ
...
-            }
+        List<Long> specificStateList = getSpecificStateList();
 
-            List<SubscriptionTimeSlot> timeSlots = null;
+        List<SubscriptionTimeSlot> timeSlots = null;
 
-            try {
-                timeSlots = subscriptionTimeSlotService.findTimeSlotsForSubscriptionsById(subscriptionIds);
+        try {
+            timeSlots = subscriptionTimeSlotService.findAllTimeSlots();
```
Kantas2601, line 614: Why looking for all Time slots here?

imi/src/main/java/org/motechproject/nms/imi/service/impl/TargetFileServiceImpl.java
```diff
@@ -577,50 +604,54 @@ private HashMap<String, Integer> generateFreshCalls(DateTime timestamp, int maxQ
...
-                    SubscriptionPack pack = subscription.getSubscriptionPack();
-                    int daysIntoPack = Days.daysBetween(subscription.getStartDate(), timestamp).getDays();
-                    if (daysIntoPack == pack.getWeeks() * 7) {
+//                    SubscriptionPack pack = subscription.getSubscriptionPack();
+                    int daysIntoPack = Days.daysBetween(new DateTime(subscription.getStartDate()), timestamp).getDays();
+//                    LOGGER.info("this is the weeks: {}",pack.getWeeks());
+                    if(subscription.getSubscriptionPackIdOid()==1){
+                        weeks =72;
+                    }else {
+                        weeks=48;
```
Kantas2601, line 652: Try to take these field from a private static final variable

kilkari/src/main/java/org/motechproject/nms/kilkari/service/impl/SubscriptionTimeSlotServiceImpl.java
```diff
@@ -27,28 +27,12 @@ public SubscriptionTimeSlotServiceImpl(SubscriptionTimeSlotDataService subscript
...
-            }
-            i++;
-        }
-
+    public List<SubscriptionTimeSlot> findAllTimeSlots() {
         String query = "SELECT ts.subscription_id, ts.subscriptionId, ts.timeStamp1, ts.timeStamp2, ts.timeStamp3 " +
-                "FROM nms_subscriptions_time_slot AS ts " +
-                "WHERE ts.subscriptionId IN (" + idsString + ")";
+                "FROM nms_subscriptions_time_slot ts " +
+                "LEFT JOIN nms_subscriptions s ON ts.subscription_id = s.id;";
```
Kantas2601, line 34: This join isn't efficient. Right now, time_slot has little data because it's in the pilot phase, but it will grow to millions soon. We might also get unnecessary old data. Add a WHERE condition to filter only recent data (e.g., last 72 weeks, exclude deactivated or completed ).

Kantas2601:
We should run this new method alongside the old way of creating the target file in production, not replace it immediately. We can't risk missing a single day's call. Use the API to generate the target file, but will not send it to Hungama. we have analyze the results first, and if everything works well, then switch to the new method completely.

@atishbeehyv123 please suggest

